PR #13, opened 2019-10-04
Fix a bug where the vat calculation is wrong when Show Net price is false

vrielsa:
Tried updating the plugin today, immediately had an issue where the tax calculated for the order wasn't being sent to Mollie (total order amount didn't match the total of the order lines). 

Even though tax_show was true and show net price false, the tax wasn't being calculated. I've fixed this in this PR and also simplified the logic a little.

### DIFF
--- a/Components/Services/BasketService.php
+++ b/Components/Services/BasketService.php
@@ -3,6 +3,8 @@
 namespace MollieShopware\Components\Services;
 
 use MollieShopware\Components\Logger;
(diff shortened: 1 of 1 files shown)


KienerNL:
Hi vrielsa,

We're looking into our tests as this wasn't spotted there. Thanks for your hard work. I will merge it right away.

My kind regards,
Reinder.

PR #15, opened 2019-10-31
Setting order sNet to hide the VAT

peter-bztrs:
When ordering from a foreign country, VAT may not be applied.
But in the order confirmation e-mail it will still be visible. By setting sNet on the Shopware Order class the VAT will not be visible in the confirmation e-mail.


KienerNL:
Thanks for the helping us out, guys!

PR #23, opened 2020-06-17
Parse order id instead of model to the getMollieOrderId() method

ArjanLankhaar:
Currently, orders cannot be shipped at Molly and you will always get the error message 'Order is paid as a single payment (not an order) at Mollie'. This will fix that.

### DIFF
--- a/Controllers/Backend/MollieOrders.php
+++ b/Controllers/Backend/MollieOrders.php
@@ -46,7 +46,7 @@ class Shopware_Controllers_Backend_MollieOrders extends Shopware_Controllers_Bac
             if (empty($order))
                 $this->returnError('Order not found');
 
-            $mollieId = $this->orderService->getMollieOrderId($order);
+            $mollieId = $this->orderService->getMollieOrderId($order->getId());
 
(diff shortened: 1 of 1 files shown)


boxblinkracer:
hi

thank you so much for this
it looked really weird for me....good thing is...it works both..because i think doctrine does already try a getId on objects

nevertheless it has to be changed
so i changed it to the other way round...so not orderId but order as paramter...because there are way more points where $order is already used :D

heres the PR
https://github.com/mollie/Shopware/pull/52

so this pr would not be necessary anymore

hope thats ok

boxblinkracer:
oh and by the way
the error you wrote is, because the order you have is done with the payments API, which means only a transaction is created in mollie and no "order"
and transactions cannot be shipped

this is due to the setting "only create orders for klarana" (yes/no)

ArjanLankhaar:
Hi @boxblinkracer,

Thanks you for your reply. If it's fixed, it's okay to me. No matter how :-p

And thanks about the tip regarding the order API configuration. I've already found that out, but thanks anyway. You can close this PR

PR #34, opened 2020-08-28
MOL-14: fix wrong mollie config for klarna shipping in subshop

boxblinkracer:
please test before merging

parts to test

* transition klarna order to a status that triggers the "shipping" in mollie
       * a) for an order from the main shop
       * b) for an order from a subshop with da different config

* ship truck icon in shopware backend
       * a) for an order from the main shop
       * b) for an order from a subshop with da different config

* refund action + partial refund
       * a) for an order from the main shop
       * b) for an order from a subshop with da different config

### DIFF
--- a/Components/Config.php
+++ b/Components/Config.php
@@ -7,6 +7,7 @@ use MollieShopware\Components\Services\ShopService;
 use Shopware\Components\Plugin\ConfigReader;
 use Shopware\Models\Order\Status;
 use Shopware\Models\Shop\Repository;
+use Shopware\Models\Shop\Shop;
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
 
 class Config
@@ -96,7 +97,7 @@ class Config
 
         // get default shop
         /** @var Repository $shopRepository */
-        $shopRepository = Shopware()->Models()->getRepository('Shopware\\Models\\Shop\\Shop');
+        $shopRepository = Shopware()->Models()->getRepository(Shop::class);
         $mainShop = $shopRepository->getActiveDefault();
 
         if ($mainShop === null) {
--- a/Components/Config.php
+++ b/Components/Config.php
@@ -5,12 +5,15 @@ namespace MollieShopware\Components;
 use Exception;
 use MollieShopware\Components\Services\ShopService;
 use Shopware\Components\Plugin\ConfigReader;
+use Shopware\Models\Order\Status;
+use Shopware\Models\Shop\Repository;
 use Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException;
 
 class Config
 {
     const TRANSACTION_NUMBER_TYPE_MOLLIE = 'mollie';
     const TRANSACTION_NUMBER_TYPE_PAYMENT_METHOD = 'payment_method';
+    const INHERITED_CONFIG_VALUE = 'inherited';
 
     /** @var ConfigReader */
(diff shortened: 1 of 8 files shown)


boxblinkracer:
achtung, offen ist noch, was passiert wenn nichts hinterlegt in subshop
hier gibt es keine konfiguration für "vererbt", müssen wir noch testen

boxblinkracer:
erledigt passt

Components/Config.php
```diff
@@ -77,6 +82,39 @@ public function get($key = null, $default = null)
...
+    private function addInheritedConfig(array $config, $name)
+    {
+        // check if inherited or null fields are in config to get from default shop
+        if (!in_array(self::INHERITED_CONFIG_VALUE, $config) && !in_array(null, $config)) {
+            return $config;
+        }
+
+        // get default shop
+        /** @var Repository $shopRepository */
+        $shopRepository = Shopware()->Models()->getRepository('Shopware\\Models\\Shop\\Shop');
```
barbieswimcrew, line 99: Shop::class instead of 'Shopware\\Models\\Shop\\Shop'?

Components/Helpers/MollieShopSwitcher.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @param $shopId
+     * @return \Mollie\Api\MollieApiClient
+     * @throws \Mollie\Api\Exceptions\ApiException
+     * @throws \Mollie\Api\Exceptions\IncompatiblePlatform
+     */
+    public function getMollieApi($shopId)
+    {
+        /** @var MollieApiFactory $apiFactory */
+        $apiFactory = Shopware()->Container()->get('mollie_shopware.api_factory');
```
barbieswimcrew, line 52: why not using $this->container?

Controllers/Backend/MollieOrders.php
```diff
@@ -189,6 +195,11 @@ public function partialRefundAction()
             if (empty($order))
                 $this->returnError('Order not found');
 
+            # switch to the config and api key of the shop from the order
```
barbieswimcrew, line 198: the following code exists three times... probably extracting it to a prif would reduce redundancy?

```
 # switch to the config and api key of the shop from the order
            $shopSwitcher = new MollieShopSwitcher($this->container);
            $this->config = $shopSwitcher->getConfig($order->getShop()->getId());
            $this->apiClient = $shopSwitcher->getMollieApi($order->getShop()->getId());
```

barbieswimcrew (changes requested):

Components/Config.php
boxblinkracer: will do tomorrow

Components/Helpers/MollieShopSwitcher.php
boxblinkracer: oh yep, copy paste error

Controllers/Backend/MollieOrders.php
boxblinkracer: this is the integration point and is also assigend different variables.
so the shop switcher is already the extracted code
also this is a small solution to avoid side effects. a better centralized access for api client and config will be done in upcoming refactorings, so for now i would leave it like this

schmitzcarsten: as it is already planned to refactor this, I will solve the issue as accepted

PR #37, opened 2020-09-16
MOL-23: Add Attributes from original Order Items to restored Basket

ThLind:
- When the basket has to be restored (because of failed payment and recreation of basket for retry) add attributes of the items to the restored Basket.

### DIFF
--- a/Components/Services/BasketService.php
+++ b/Components/Services/BasketService.php
@@ -2,31 +2,43 @@
 
 namespace MollieShopware\Components\Services;
 
+use Enlight_Components_Db_Adapter_Pdo_Mysql;
+use Exception;
+use MollieShopware\Components\Config;
 use MollieShopware\Components\Logger;
-use Shopware\Models\Article\Detail;
+use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Order\Basket;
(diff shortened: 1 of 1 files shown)


Components/Services/BasketService.php
```diff
@@ -37,22 +49,23 @@ public function __construct(\Shopware\Components\Model\ModelManager $modelManage
 
         $this->orderService = Shopware()->Container()
             ->get('mollie_shopware.order_service');
+
+        $this->db = Shopware()->Db();
```
barbieswimcrew, line 53: Shopware()->Container()->get('db') possibly?

Components/Services/BasketService.php
```diff
@@ -376,7 +473,7 @@ public function resetOrderDetailQuantity(\Shopware\Models\Order\Detail $orderDet
             try {
                 $this->modelManager->persist($article);
                 $this->modelManager->flush($article);
-            } catch(\Exception $e) {
+            } catch(Exception $e) {
```
barbieswimcrew, line 476: please format code

barbieswimcrew (changes requested):

barbieswimcrew (approved):

PR #39, opened 2020-09-21
MOL-14: use sub shop config in klarna cli command

boxblinkracer:
switching of config and api client was missing in cli command

added it and refactored the if conditions a bit with early returns

### DIFF
--- a/Command/KlarnaShippingCommand.php
+++ b/Command/KlarnaShippingCommand.php
@@ -5,6 +5,8 @@ namespace MollieShopware\Command;
 use Mollie\Api\MollieApiClient;
 use MollieShopware\Components\Config;
 use MollieShopware\Components\Constants\PaymentMethod;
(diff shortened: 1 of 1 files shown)


Command/KlarnaShippingCommand.php
```diff
@@ -63,53 +65,75 @@ public function execute(InputInterface $input, OutputInterface $output)
...
+
+                    /** @var Order $order */
+                    $order = $orderRepository->find($transaction->getOrderId());
+
+                    if ($order === null) {
+                        continue;
+                    }
+
+
+                    $switcher = new MollieShopSwitcher($this->container);
```
barbieswimcrew, line 101: Will it actually be necessary to (re)create that newable at any of the foreach loop's iterations? Or may the switcher instantiation be placed outside the loop?

barbieswimcrew (changes requested):

barbieswimcrew (approved):

PR #40, opened 2020-09-21
NTR: improve notification endpoint exception handling

boxblinkracer:
due to the new feature of providing a test mode with both, live and api key in the configuration,
there is a new acceptance test that proves the usage of the subshop config in the notification endpoint

however i found out that the endpoint is not really stable. shopware error in case of throwables instead of json response.
also errors where displayed to the requester which might not be good to see ;)

so i've added a try catch with throwable catching
along with better logging and a generic error output

### DIFF
--- a/Controllers/Frontend/Mollie.php
+++ b/Controllers/Frontend/Mollie.php
@@ -337,25 +337,28 @@ class Shopware_Controllers_Frontend_Mollie extends AbstractPaymentController
      */
     public function notifyAction()
     {
-        /** @var \Shopware\Models\Order\Order $order */
-        $order = null;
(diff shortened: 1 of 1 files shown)


Controllers/Frontend/Mollie.php
```diff
@@ -387,11 +390,22 @@ public function notifyAction()
                     'Order not found'
                 );
             }
-        } catch (\Exception $e) {
-            Notifier::notifyException(
-                $e->getMessage(),
-                $e
+        } catch (\Trhowable $e) {
```
barbieswimcrew, line 393: typo

barbieswimcrew (changes requested):

barbieswimcrew (approved):

PR #51, opened 2020-09-25
MOL-30: Hide original Apple Pay on payment selection page

boxblinkracer:
the original apple pay was hidden, even its a default payment method...no one knew...that was the function

i've also renamed the function and improved descriptions

### DIFF
--- a/Resources/views/frontend/_public/src/js/applepay.js
+++ b/Resources/views/frontend/_public/src/js/applepay.js
@@ -1,0 +1,49 @@
+(function ($) {
+    "use strict";
+
+    $(document).ready(function() {
+        hideApplePayIfNotAllowed();
+    });
+
+    $(document).ajaxComplete(function() {
+        hideApplePayIfNotAllowed();
+    });
+
+    /**
+     * This function is used to hide Apple Pay if the visitor can't use it.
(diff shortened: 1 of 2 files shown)


Resources/views/frontend/_public/src/js/applepay.js
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    $(document).ajaxComplete(function() {
+        hideApplePayIfNotAllowed();
+    });
+
+    /**
+     * This function is used to hide Apple Pay if the visitor can't use it.
+     * Please keep in mind, this is Apple Pay and not Apple Pay Direct.
+     * So it just hides Apple Pay on the payment selection page.
```
barbieswimcrew, line 15: wouldn't it be better to run all this only on checkout/payment (don't know the exact controller action related css class of body tag but there might be a specific one to distinguish)? otherwise that js code will be fired on any page's load and even on any ajaxComplete event... please think about that

barbieswimcrew (changes requested):

Resources/views/frontend/_public/src/js/applepay.js
boxblinkracer: yes in refactoring, for now the prio is to revert that functionality to have it back

PR #64, opened 2020-10-09
MOL-21: Return CreditCardToken

ThLind:
In CreditCartService a card token has been generated for B2B but never returned. Added return statement.


boxblinkracer:
nice one

PR #66, opened 2020-10-19
MOL-41: Apple Pay Direct with logged in Shop Account

boxblinkracer:
if the user is already signed in, apple pay direct should use this user instead of creating a guest account.

the user data by apple pay is now stored in the session.

if required, a guest account is being created.

however in both cases, it will be ensured that the order will get the data from apple pay

### DIFF
--- a/Components/Account/Account.php
+++ b/Components/Account/Account.php
@@ -61,6 +61,18 @@ class Account
     }
 
     /**
+     * Gets if the user is already signed in.
+     * 
+     * @return bool
+     */
+    public function isLoggedIn()
+    {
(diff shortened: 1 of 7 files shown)


boxblinkracer:
on hold (after new release) @barbieswimcrew

barbieswimcrew (approved):

barbieswimcrew:
@boxblinkracer approved, please merge on your own if the right time has come ;-)

boxblinkracer:
bedankt voor alles @barbieswimcrew  ;)

PR #74, opened 2020-10-26
MOL-54: improve configuration for custom webhooks urls

boxblinkracer:
more easy config, refactored + separate class

### DIFF
--- a/Components/CustomConfig/CustomConfig.php
+++ b/Components/CustomConfig/CustomConfig.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace MollieShopware\Components\CustomConfig;
+
+class CustomConfig
+{
+
+    /**
+     * main node for a custom mollie configuration
+     */
+    const KEY_MOLLIE_CONFIG = 'mollie';
+
+    /**
+     * @var array
+     */
+    private $config;
+
+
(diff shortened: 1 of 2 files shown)


Components/CustomConfig/CustomConfig.php
```diff
@@ -0,0 +1,54 @@
...
+    public function __construct($shopwareConfig)
+    {
+        # let's verify if we have a custom config for mollie
+        # with that custom config its possible to use a different
+        # options for developers that might not be possible in 
+        # the default plugin
+        if (isset($shopwareConfig[self::KEY_MOLLIE_CONFIG])) {
+            $this->config = $shopwareConfig[self::KEY_MOLLIE_CONFIG];
+        } else {
+            $this->config = array();
```
barbieswimcrew, line 33: why not declaring the empty array in line 16 when defining the $config class param?

barbieswimcrew (approved):

Components/CustomConfig/CustomConfig.php
boxblinkracer: no fan ob decentralised assignment in this case
because the constructor would only assign 50% of the truth :)

PR #79, opened 2020-10-27
MOL-60: Add Unit Tests + Travis CI

boxblinkracer:
finally added phpunit and tests

attention:
its a dev dependency. unit tests work only in php 7.x and not 5.6
due to some syntax changes in phpunit

so we always ship the plugin without DEV dependencies, to have it compatible for php 5.6

Travis CI pipe has been improved.
php 5.6 and 7.0+7.1 (composer problems) will not run unit tests, only syntax checks.
all other php versions do finally run unit tests too


boxblinkracer:
@barbieswimcrew please see travis build pipes and verify

from that point on, merges are only allowed on successful pipes (finally....rrrrrr) :D

barbieswimcrew (approved):

PR #88, opened 2020-11-11
MOL-74: Fix broken payment installation without Live API key

boxblinkracer:
somehow getActivePaymentMethodsFromMollie is null
this seems to be due to a missing api key

i think there is something weird anyway in the installation process
i'just tried to avoid that exception, then it works and just doesnt install anything

could you double check @barbieswimcrew  (that was your code - just to be sure)

### DIFF
--- a/Components/Services/PaymentMethodService.php
+++ b/Components/Services/PaymentMethodService.php
@@ -111,7 +111,15 @@ class PaymentMethodService
         $position = 0;
 
         /** @var null|MethodCollection $methods */
-        $methods = $this->appendApplePayDirectFeature($this->getActivePaymentMethodsFromMollie());
+        $methods = $this->getActivePaymentMethodsFromMollie();
+
+        if ($methods !== null) {
+            # if its not null, do the same again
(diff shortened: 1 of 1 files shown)


Components/Services/PaymentMethodService.php
```diff
@@ -111,7 +111,15 @@ public function getPaymentMethodsFromMollie()
         $position = 0;
 
         /** @var null|MethodCollection $methods */
-        $methods = $this->appendApplePayDirectFeature($this->getActivePaymentMethodsFromMollie());
+        $methods = $this->getActivePaymentMethodsFromMollie();
+
+        if ($methods != null) {
```
barbieswimcrew, line 116: !==

barbieswimcrew (changes requested):

Components/Services/PaymentMethodService.php
boxblinkracer: erledigt, danke

barbieswimcrew (approved):

PR #91, opened 2020-11-17
MOL-75: Refactoring of line item rounding

boxblinkracer:
ON HOLD

### DIFF
--- a/Components/Basket/Basket.php
+++ b/Components/Basket/Basket.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace MollieShopware\Components\Basket;
+
+use MollieShopware\Components\TransactionBuilder\Models\BasketItem;
+use Psr\Log\LoggerInterface;
+use Shopware\Components\Model\ModelManager;
+use Shopware\Models\Order\Repository;
+
+class Basket
+{
+
+    /**
+     * @var ModelManager $modelManager
+     */
+    private $modelManager;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    
+    /**
+     * Basket constructor.
+     * @param ModelManager $modelManager
+     * @param LoggerInterface $logger
+     */
+    public function __construct(ModelManager $modelManager, LoggerInterface $logger)
+    {
+        $this->modelManager = $modelManager;
+        $this->logger = $logger;
+    }
+
+
+    /**
+     * Get positions from basket
+     *
+     * @param array $userData
+     * @return BasketItem[]
+     * @throws \Exception
+     */
+    public function getBasketLines($userData = array())
+    {
+        $items = [];
+
+        try {
+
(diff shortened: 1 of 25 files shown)


boxblinkracer:
ON HOLD, is in preparation status

boxblinkracer:
The rounding was wrong for line items when they are sent to mollie.
these need to be 100% accurate (vat amount within each line, etc.)


**Basic Workflow**
the basket items along with a shipping lines will be calculated and added as transaction items to the database.
these items will be sent to mollie as line items, so they have to be calculated correctly.
and that was broken

**what i did?!**
i moved both, the transaction item builder (for database) and request-json-array builder (for mollie request) functions to separate testable classes.

the calculation is now done correctly within the TransactionItemBuilder.

i've added some refactoring to make shipping costs work exactly like basket items too.
and then they will use the transaction item builder with the correct rounding too

i did also add some more tests, code coverage to make sure its tested, and some adjustments for travis ci (remove test analytics due to php 5.6 and :void returns)

boxblinkracer:
GO for review @barbieswimcrew

Components/Services/PaymentService.php
```diff
@@ -478,16 +476,17 @@ private function prepareRequest(
             'method' => $paymentMethod,
         ];
 
-        $paymentDescription = (string) (time() . $transaction->getId() . substr($transaction->getBasketSignature(), -4));
+        $paymentDescription = (string)(time() . $transaction->getId() . substr($transaction->getBasketSignature(), -4));
```
barbieswimcrew, line 479: -4 as constant?

barbieswimcrew (changes requested):

Components/Basket/Basket.php
```diff
@@ -0,0 +1,97 @@
...
+                );
+               
+                # update our basket item ID if we have that attribute
+                # i dont know why - isn't it in the ID already?, but let's keep with it for now
+                if ($basketItem !== null && $basketItem->getAttribute() !== null && method_exists($basketItem->getAttribute(), 'setBasketItemId')) {
+
+                    $basketItem->getAttribute()->setBasketItemId($basketItem->getId());
+
+                    $this->modelManager->persist($basketItem);
+                    $this->modelManager->flush($basketItem);
```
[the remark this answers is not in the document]
boxblinkracer: that is no change it was always like this and i rather stick with this so far because i dont even know why its done haha

Components/Services/PaymentService.php
boxblinkracer: haha, i can see what the hell it even is :D

boxblinkracer: ok i have no clue what -4 is , and dont feel good in changing it in this PR :)

barbieswimcrew (approved):

PR #92, opened 2020-11-20
MOL-71: Compatibility Fix for empty payment ID

boxblinkracer:
sometimes the payment ID is somehow NULL and sSaveOrder throws an exception

that means that that order will not be saved, which is indeed a problem

we have no clue why the paymentID is somehow NULL...maybe hooks, events or other plugins?

so to fix this, the paymentID will be kept in a separate session key.
if its somehow NULL, we just restore that paymentID and set it again in our session
(worked while simulating it)

### DIFF
--- a/Controllers/Frontend/Mollie.php
+++ b/Controllers/Frontend/Mollie.php
@@ -21,6 +21,7 @@ use MollieShopware\Components\TransactionBuilder\Models\TaxMode;
 use MollieShopware\Components\TransactionBuilder\TransactionItemBuilder;
 use MollieShopware\Models\Transaction;
 use MollieShopware\Models\TransactionRepository;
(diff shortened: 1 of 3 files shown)


Controllers/Frontend/Mollie.php
```diff
@@ -308,55 +318,69 @@ public function returnAction()
...
 
-        if (
-            $transactionNumber !== ''
-            && ($order === null || !$order instanceof \Shopware\Models\Order\Order)
-        ) {
-            $order = $this->getOrderFromTransaction($transactionNumber);
-        }
+        try {
 
-        if ($order === null) {
```
boxblinkracer, line 318: looks worse than it is :)
just wrapped it with a try catch

barbieswimcrew (approved):

PR #97, opened 2020-11-27
ON-HOLD: MOL-68: Prevent ApplePayDirect as retry Method on Confirm Page

ThLind:
When an Error occurs during Apple Pay Direct Payment the customer is redirected to /checkout/confirm Page which display an Error Message and offers the opportunity to retry placing the order. When on this Page Apple Pay Direct is selected as Payment Method it didn't work as this has to be started by the Apple Pay Direct Button. 
So as a Solution prevent that Apple Pay Direct is selected as retry Payment Method. Instead if it's selected switch to Apple Pay so that the order can be completed.

### DIFF
--- a/Subscriber/ApplePayDirectSubscriber.php
+++ b/Subscriber/ApplePayDirectSubscriber.php
@@ -88,12 +88,19 @@ class ApplePayDirectSubscriber implements SubscriberInterface
      */
     public function onFrontendCheckoutPostDispatch(Enlight_Event_EventArgs $args)
     {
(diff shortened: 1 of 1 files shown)


boxblinkracer:
not working for me
the failed order does always have Vorkasse on confirm, so that code will never be triggered

boxblinkracer (changes requested):
not working for me.
the failed order on the confirm will always have Vorkasse for me
so that code will never be triggered

boxblinkracer:
is now merged in another PR

PR #100, opened 2020-12-10
MOL-92: Add Index if not exists

ThLind:
Add Index to mol_sw_transactions on activate if the Index doesn't exist already.

### DIFF
--- a/MollieShopware.php
+++ b/MollieShopware.php
@@ -260,6 +260,9 @@ class MollieShopware extends Plugin
         $downloader = new ApplePayDomainFileDownloader();
         $downloader->downloadDomainAssociationFile(Shopware()->DocPath());
 
+        // add index to mol_sw_transactions if not exists
+        $this->addIndexToTransactions();
+
         parent::activate($context);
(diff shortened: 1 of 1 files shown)


MollieShopware.php
```diff
@@ -533,4 +536,15 @@ private function cleanBackendSnippets()
         $cleaner->cleanBackendSnippets();
     }
 
+    private function addIndexToTransactions()
+    {
+        $connection = $this->container->get('dbal_connection');
+
+        $indexExistsCheck = $connection->executeQuery("SELECT COUNT(1) indexIsThere FROM INFORMATION_SCHEMA.STATISTICS WHERE table_schema=DATABASE() AND table_name='mol_sw_transactions' AND index_name='transaction_id_idx';")->fetch();
+
+        if ($indexExistsCheck['indexIsThere'] == 0) {
```
barbieswimcrew, line 545: This condition could probably be typesafe? ===

barbieswimcrew (changes requested):

PR #101, opened 2021-01-07
MOL-103: Refactoring Payment Flows

boxblinkracer:
as discussed on the phone

### DIFF
--- a/Components/Config.php
+++ b/Components/Config.php
@@ -205,11 +205,15 @@ class Config
         /** @var int|null $level */
         $level = $this->get('log_level');
 
+        if ($level === '2, INFO') {
+            $level = Logger::INFO;
+        }
+
         if ($level === null) {
             return Logger::INFO;
         }
 
-        return $level;
+        return (int)$level;
     }
 
     /**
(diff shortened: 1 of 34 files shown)


Components/Config.php
```diff
@@ -236,7 +240,7 @@ public function getBankTransferDueDateDays()
      *
      * @return boolean
      */
-    public function sendStatusMail()
+    public function sendPaymentStatusChangedMail()
```
schmitzcarsten, line 243: is this method name good enough. You return a bbolean if a mail should be sent. The method name means that a mail is sent here.

Facades/CheckoutSession/CheckoutSessionFacade.php
```diff
@@ -0,0 +1,300 @@
...
+            # this is required to restore the basket from that order in case of
+            # any additional failures below. otherwise the cart is empty because
+            # the order is already completed.
+            $this->restorableOrder = $order;
+
+            # now update all required references of our transaction
+            $transaction->setOrderId($order->getId());
+            $transaction->setOrderNumber($orderNumber);
+
+            # todo fehlt hier flush?
```
schmitzcarsten, line 207: todo found

schmitzcarsten (commented):
added two little issues

Facades/CheckoutSession/CheckoutSessionFacade.php
boxblinkracer: jap, checking that while testing today, thx

PR #103, opened 2021-01-15
MOL-117: Add anonymized payment token to logs

boxblinkracer:
we sometimes have to debug i payment tokens are correctly used and existing
so it will now add tehese to the initial log entry

its of course anonymized and tested


Controllers/Frontend/Mollie.php
```diff
@@ -458,6 +460,8 @@ private function loadServices()
             $orderUpdater = new OrderUpdater($this->config, $sOrder);
             $confirmationMail = new ConfirmationMail($sOrder, $repoTransactions);
 
+            $tokeAnonymizer = new TokenAnonymizer('*', 4, 15);
```
barbieswimcrew, line 463: I'd love to see those numbers being constants 🔢

barbieswimcrew (changes requested):

barbieswimcrew (approved):

PR #104, opened 2021-01-15
MOL-100: Cleanup ordermailVariables and unset them

ThLind:
Move unsetting of ordermail variables to return action to also unset them when payment failed.

Add cleanup to activate action of plugin to clean up old ordermail variables.

### DIFF
--- a/MollieShopware.php
+++ b/MollieShopware.php
@@ -2,6 +2,7 @@
 
 namespace MollieShopware;
 
+use Doctrine\DBAL\Driver\Connection;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Enlight_Template_Manager;
 use Exception;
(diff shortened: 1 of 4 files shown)


MollieShopware.php
```diff
@@ -547,4 +552,11 @@ private function addIndexToTransactions()
         }
     }
 
+    private function cleanOrdermailVariables()
+    {
+        /** @var Connection $connection */
+        $connection = $this->container->get('dbal_connection');
+
+        $connection->executeQuery('UPDATE mol_sw_transactions set ordermail_variables = NULL');
```
barbieswimcrew, line 560: just a little something... i'd prefer to see the SET keyword in uppercase as well as the UPDATE already is

barbieswimcrew (approved):

PR #115, opened 2021-01-21
[MOL-97] Refund from CLI

Sironheart:


barbieswimcrew (approved):

Command/OrdersRefundCommand.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if(\is_array($orderNumber) ||
+            (
+                $refundAmount !== null &&
+                \is_array($refundAmount) &&
+                !\is_numeric($refundAmount)
+            )
+        ) {
+            throw new InvalidArgumentException(
+                'There was an error during the input of information. Please only submit one orderNumber per execution and set refund amounts to be split with a dot.',
```
boxblinkracer, line 93: what would be that number of amounts?
amount line items? how would you be able to detect and match line items - theres no order in the items?

Command/OrdersRefundCommand.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * @param string|array $orderNumber
+     * @param string|array|null $refundAmount
+     *
+     * @throws InvalidArgumentException
+     */
+    private function validateInputArguments($orderNumber, $refundAmount)
+    {
+        if(\is_array($orderNumber) ||
```
boxblinkracer, line 85: why array? isnt that just one number?

Components/Services/OrderService.php
```diff
@@ -95,6 +99,28 @@ public function getOrderDetailById($orderDetailId)
...
+     * @param string $orderNumber
+     *
+     * @return Transaction
+     *
+     * @throws TransactionNotFoundException
+     */
+    public function getOrderTransactionByNumber($orderNumber)
+    {
+        /** @var Repository $transactionRepository */
+        $transactionRepository = $this->modelManager->getRepository(Transaction::class);
```
boxblinkracer, line 112: somehow i feel i've seen a couple of these methods?
could you double check, if its not existing its totally fine

Controllers/Backend/MollieOrders.php
```diff
@@ -65,16 +75,21 @@ public function shipAction()
             $mollieOrder = $this->apiClient->orders->get($mollieId);
             $errorMessage = '';
 
-            if (empty($mollieOrder))
+            if ($mollieOrder === null) {
```
boxblinkracer, line 78: thx for touching that code. makes it better. you could indeed do an "instance of Order" check to even increase it a bit more if you want

Resources/services.xml
```diff
@@ -40,6 +40,12 @@
             <argument type="service" id="mollie_shopware.components.logger"/>
         </service>
 
+        <service class="MollieShopware\Command\OrdersRefundCommand">
+            <tag name="console.command" command="mollie:orders:refund"/>
```
boxblinkracer, line 44: do we have the command duplicated in the php and xml? just thinking about what is used :D 
maybe you can just do it in 1 place..(i think in php for consistency?)

Services/RefundInterface.php
```diff
@@ -0,0 +1,29 @@
...
+use Mollie\Api\Resources\Order as MollieOrder;
+use Mollie\Api\Resources\OrderLine;
+use Mollie\Api\Resources\Payment as MolliePayment;
+use MollieShopware\Models\Transaction;
+use Shopware\Models\Order\Detail;
+use Shopware\Models\Order\Order;
+
+interface RefundInterface
+{
+    public function refundOrderAmount(Order $order, Transaction $transaction, float $customAmount = null);
```
boxblinkracer, line 16: just a big thank you for an interface :)

question though:
is this the old implementation or your new one?
it seems as if some things our duplicated like processRefund...wouldnt that be more a private thing because we already have a few public refunds?
also i'd be happy if we could remove optional params such as $quantity=1, $customAmount=null, it easily leads to errors and different behaviours of a method.
...in case its really a new code, if its an old one we can keep it for now to avoid breaking anything

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+    private $modelManager;
+
+    public function __construct(EntityManager $modelManager)
+    {
+        $this->modelManager = $modelManager;
+    }
+
+    public function refundOrderAmount(Order $order, Transaction $transaction, float $customAmount = null)
+    {
+        $mollieClient = $this->getMollieApi($order->getShop()->getId());
```
boxblinkracer, line 35: in the code, there is a MollieShopSwitcher which was introduced as some small helper because of a problem with sub shops....
could you double check that we use the correct plugin configuration (api keys, etc...) of that particular order.
if that order is from subshop B, it could be that subshop B has a different API key...just to make sure its working thank you

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+
+        if ($mollieClient === null) {
+            throw new \Exception('Something went wrong trying to get an API Client Instance. Please try again later.', 1);
+        }
+
+        if ($transaction->getMolliePaymentId() !== null) {
```
boxblinkracer, line 41: i think $transaction should have something like "isOrderType" or such a getter
could you use that for these ifs?

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+            $molliePayment = $mollieClient->payments->get($transaction->getMolliePaymentId());
+
+            if (
+                (
+                    $customAmount === null ||
```
boxblinkracer, line 46: could we create unit tests for all these checks if a refund is allowed or not?
maybe separate validator + unit tests and then just use it in here?

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * @return bool|Refund
+     * @throws \Exception
+     *
+     */
+    public function refundOrder(Order $order, MollieOrder $mollieOrder)
+    {
+        $apiClient = $this->getMollieApi($order->getShop()->getId());
+
+        if (empty($this->modelManager) || $apiClient === null) {
```
boxblinkracer, line 89: if it works with fail-fast just remove it, or add an exception if the if is required, thx

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+    public function partialRefundOrder(
+        Order $order,
+        Detail $detail,
+        MollieOrder $mollieOrder,
+        OrderLine $orderLine,
+        $quantity = 1
+    ) {
+        $apiClient = $this->getMollieApi($order->getShop()->getId());
+
+        if (empty($this->modelManager) || $apiClient === null) {
```
boxblinkracer, line 138: if it works with fail-fast just remove it, or add an exception if the if is required, thx

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+                    'id' => $orderLine->id,
+                    'quantity' => $quantity,
+                ],
+            ]
+        ];
+
+        /** @var Refund $refund */
+        $refund = $apiClient->orderRefunds->createFor($mollieOrder, $data);
+
+        if ($refund !== null) {
```
boxblinkracer, line 154: shouldnt there be an exception in here?
or is it totally fine that null will be returned? :)

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+        $refund = $molliePayment->refund(
+            [
+                'amount' => [
+                    'currency' => $order->getCurrency(),
+                    'value' => number_format($order->getInvoiceAmount(), 2, '.', '')
+                ]
+            ]
+        );
+
+        if ($refund !== null) {
```
boxblinkracer, line 183: same here with null return?

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+    public function partialRefundPayment(Order $order, MolliePayment $molliePayment, $amountToRefund)
+    {
+        $refund = $molliePayment->refund([
+            'amount' => [
+                'currency' => $order->getCurrency(),
+                'value' => number_format($amountToRefund, 2, '.', '')
+            ]
+        ]);
+
+        if ($refund !== null) {
```
boxblinkracer, line 210: null return is ok?

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * @param Order $order
+     *
+     * @return bool
+     *
+     * @throws \Exception
+     */
+    public function processRefund(Order $order)
+    {
+        if (empty($this->config) || empty($this->modelManager)) {
```
boxblinkracer, line 228: if it works with fail-fast just remove it, or add an exception if the if is required, thx

Services/RefundService.php
```diff
@@ -0,0 +1,281 @@
...
+            $mollieReturn = $detail->getAttribute()->getMollieReturn();
+            $mollieReturn += $quantity;
+
+            $detail->getAttribute()->setMollieReturn($mollieReturn);
+
+            try {
+                $this->modelManager->persist($detail->getAttribute());
+                $this->modelManager->flush($detail->getAttribute());
+            } catch (Exception $e) {
+                //
```
boxblinkracer, line 277: should we remote the try catch?
it might fail without any hint?

boxblinkracer (changes requested):
see comments, thx

Services/RefundInterface.php
Sironheart: it's the old code from the backend controller, I think :)

Command/OrdersRefundCommand.php
Sironheart: this is a basic check, because symfony can take different things here. Arrays, strings, nothing at all, etc... these are just the ones which might occur in a normal use case :)

Controllers/Backend/MollieOrders.php
Sironheart: I'd prefer to leave this as a null check. instanceof makes the code harder to read and should always be returned by the `$this->apiClient->get()` method

Services/RefundService.php
Sironheart: I could add that. But since we only need the IDs I think this would be a small overkill, isn't it?

Command/OrdersRefundCommand.php
Sironheart: I think partial order refunds would need an additional command, since the current concept does not quite consider partial order refunds :/ I believe it would be best to move partial order refunds into their own command, since this just differs too much, in terms of the inputs, from the other refunds :/

Components/Services/OrderService.php
Sironheart: image

These are all occurences I found. I believe every single one of these occurences differ a bit.

Controllers/Backend/MollieOrders.php
boxblinkracer: i generally prefere comparing to what we actually need and not that its not just null.
i cannot agree with harder to read, for me its a better stability in the end, but its ok you can leave it

Services/RefundService.php
boxblinkracer: its just a getter in your existing object that does those checks internally.
i try to move to an internal oop approach instead of doing does checks outside
https://github.com/mollie/Shopware/blob/master/Models/Transaction.php#L432

Components/Services/OrderService.php
boxblinkracer: sorry i meant the whole new created function to get an order by transaction number.
just checked it, i mixed it with getOrderByNumber, this is getTransactionByNumer, all good

Command/OrdersRefundCommand.php
boxblinkracer: i thought we talked about a an optional --full command that could use?
i rather avoid having 2 commands for 1 feature, thats why options and arguments exist
and the entrypoint which triggers based on e.g. an argument has nothing to do with the internally used services.
so for me you can do beautiful reliable internal services for full or partial, but please use 1 entrypoint command

PR #116, opened 2021-01-21
[MOL-107] Adding a payment method import button to the plugin configuration

Sironheart:


Resources/config.xml
```diff
@@ -122,6 +122,33 @@
             </options>
         </element>
 
+        <element type="button">
+            <name>btnPaymentUpdate</name>
+            <label lang="en">update Payment methods</label>
+            <label lang="de">Zahlungsmethoden aktualisieren</label>
```
boxblinkracer, line 128: hi
niederländisch wäre: "Betaalmethoden actualiseren"

Resources/config.xml
```diff
@@ -122,6 +122,33 @@
             </options>
         </element>
 
+        <element type="button">
+            <name>btnPaymentUpdate</name>
```
boxblinkracer, line 126: please add the position entry.
this is mandatory to always keep a consisent plugin view for new installations and updates

Controllers/Backend/MolliePayments.php
```diff
@@ -0,0 +1,55 @@
...
+                $methods = $paymentMethodService->getPaymentMethodsFromMollie();
+            }
+
+            // Install the payment methods from Mollie
+            if ($methods !== null) {
+                $paymentMethodService->installPaymentMethod($this->container->getParameter('mollie_shopware.plugin_name'), $methods);
+            }
+
+            // download apple pay merchant domain verification file of mollie
+            $downloader = new ApplePayDomainFileDownloader();
```
boxblinkracer, line 36: that should not happen in here,
that just downloads the domain verification file when installing and updating the plugin

Controllers/Backend/MolliePayments.php
```diff
@@ -0,0 +1,55 @@
...
+            $this->View()->assign('response', $e->getMessage());
+        }
+    }
+
+    /**
+     * @return LoggerInterface
+     */
+    private function getPluginLogger()
+    {
+        return $this->container->get('pluginlogger');
```
boxblinkracer, line 53: please use dedicated mollie logger and not the plugin logger

Controllers/Backend/MolliePayments.php
```diff
@@ -0,0 +1,55 @@
...
+            // Install the payment methods from Mollie
+            if ($methods !== null) {
+                $paymentMethodService->installPaymentMethod($this->container->getParameter('mollie_shopware.plugin_name'), $methods);
+            }
+
+            // download apple pay merchant domain verification file of mollie
+            $downloader = new ApplePayDomainFileDownloader();
+            $downloader->downloadDomainAssociationFile(Shopware()->DocPath());
+
+            Shopware()->Template()->assign('response', 'Success!');
```
boxblinkracer, line 39: please show something for the simple merchant user like
"xy Payment Methods imported/updated" .....
simple echo without template is fine, or inline html echo...as we've discussed in kickoff

Controllers/Backend/MolliePayments.php
```diff
@@ -0,0 +1,55 @@
...
+use Shopware\Components\Plugin\ConfigReader;
+use Shopware\Components\Plugin\PaymentInstaller;
+
+class Shopware_Controllers_Backend_MolliePayments extends Shopware_Controllers_Backend_ExtJs
+{
+    public function updateAction()
+    {
+        try {
+            /** @var PaymentMethodService $paymentMethodService */
+            $paymentMethodService = $this->container->get('mollie_shopware.payment_method_service');
```
boxblinkracer, line 20: please use mollie logger with 
INFO: Importing Payment Methods
....
and ERROR in catch

Controllers/Backend/MolliePayments.php
```diff
@@ -0,0 +1,55 @@
...
+
+            if ($paymentMethodService !== null) {
+                // Deactivate all Mollie payment methods
+                $paymentMethodService->deactivatePaymentMethods();
```
boxblinkracer, line 24: why are you deactivating molie payments in here?
shouldnt this only be done when deactivating the plugin? 
what if this happens accidentally. its not the place where one would expect this to happen

boxblinkracer (changes requested):
please see comments. danke :)

Controllers/Backend/MolliePayments.php
Sironheart: During the request Mollie gets asked which payments are available... when the payments aren't available anymore it should be disabled. But since the paymentMethodService does activate them this should only be a 'problem', if there already IS a problem with this payment method

boxblinkracer: good one, totally fine for me
So we have an auto-deactivation feature, gfoit ma :) danke

Controllers/Backend/MolliePayments.php
Sironheart: I would rather use the Template engine here, because this is the way shopware itself does it. Any other way would end in possible compatability problems.

boxblinkracer (approved):

barbieswimcrew (approved):

PR #119, opened 2021-01-26
Mol 129/fallback shipping costs for direct payments

ThLind:
In case of Direct Payments a Fallback has been added which reads the ShippingCost Settings from Backend and calculates the Shipping Costs.

This PR is based on MOL-76

### DIFF
--- a/Components/Shipping/Providers/CartShippingCostsProvider.php
+++ b/Components/Shipping/Providers/CartShippingCostsProvider.php
@@ -2,18 +2,32 @@
 
 namespace MollieShopware\Components\Shipping\Providers;
 
-use MollieShopware\Components\Shipping\ShippingCostsProviderInterface;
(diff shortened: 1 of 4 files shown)


barbieswimcrew (approved):

Components/Shipping/Providers/CartShippingCostsProvider.php
```diff
@@ -32,4 +54,50 @@ public function getShippingCosts()
         );
     }
 
+    /**
+     * @return array|null
+     * @throws OrderNotFoundBySessionIdException
+     */
+    private function getInvoiceShippingCosts()
+    {
+        $sessionId = Shopware()->Session()->offsetGet('sessionId');
```
boxblinkracer, line 63: can you move session to DI?

PR #120, opened 2021-01-27
[MOL-57] Mark transactions as shipped when shipped via backend.

Sironheart:


Components/Services/PaymentService.php
```diff
@@ -784,6 +787,10 @@ public function sendOrder($mollieId)
         try {
             /** @var \Mollie\Api\Resources\Order $mollieOrder */
             $mollieOrder = $this->apiClient->orders->get($mollieId);
+
+            if ($mollieOrder->isShipping()) {
+                return (new ArrayCollection($mollieOrder->shipments()->getArrayCopy()))->last();
```
boxblinkracer, line 792: hi, just to verify this...the code below with saveTransactionAsShipped looks good, but does this?

Controllers/Backend/MollieOrders.php
```diff
@@ -59,30 +60,52 @@ public function shipAction()
...
                 $errorMessage = 'The order is already shipping at Mollie.';
+            }
 
             if ((string) $errorMessage !== '') {
                 $this->returnError($errorMessage);
             }
 
             $result = $mollieOrder->shipAll();
 
+            if ($this->modelManager === null) {
```
boxblinkracer, line 92: i think we should remove that null? it has to exist anyway, right?

barbieswimcrew (approved):

Components/Services/PaymentService.php
Sironheart: this is basicly only returning the last shipment, that gets found, because you can not ship things, that are already shipped, I'd say

Sironheart:
Merge Conflicts will be resolved on approval :)

Components/Services/PaymentService.php
boxblinkracer: ah ok and shippings get returned..ok :)

boxblinkracer (approved):

PR #139, opened 2021-02-09
MOL-69: Mollie User in Order History

Sironheart:
Also some dead code removal


barbieswimcrew (approved):

boxblinkracer:
info -> must not be merged with this approach. well be changed after our meeting

StephanAltmann85:
> info -> must not be merged with this approach. well be changed after our meeting

Pull request updated

Subscriber/OrderBackendSubscriber.php
```diff
@@ -156,6 +161,39 @@ public function onOrderPostDispatch(Enlight_Controller_ActionEventArgs $args)
...
+            $view->assign([
+                'success' => false,
+                'data' => $this->Request()->getParams(),
+                'message' => $namespace->get('no_order_id_passed', 'No valid order id passed.'),
+            ]);
+
+            return;
+        }
+
+        $history = $this->orderService->getOrderStatusHistory($orderId, $sort, $offset, $limit);
```
boxblinkracer, line 189: short question, if we just extend the frontend, why do we need all of this?
just to add the "comments" column? is there another way besides having that bigger repo criteria query "duplicated"?

Facades/Notifications/Notifications.php
```diff
@@ -175,7 +181,8 @@ public function onNotify($transactionID, $paymentID)
             }
         }
 
+        $this->orderUpdater->updateOrderHistoryComment($order);
```
boxblinkracer, line 184: doesnt this this need to be somewhere deeper integrated?
if not, then i think its missing in the "returnAction"....there are also status updates when the user comes back
so 1x return + 1x notification .......and probably in extjs? just thinking about what happens in shipping actions there

Components/Services/OrderService.php
```diff
@@ -179,6 +180,41 @@ public function getMollieOrderId($order)
...
     }
 
+    /**
+     * @param int $orderId
+     * @param array|null $orderBy
+     * @param int|null $offset
+     * @param int|null $limit
+     * @return array|int|string
+     */
+    public function getOrderStatusHistory($orderId, $orderBy, $offset, $limit) {
```
boxblinkracer, line 190: this is the question from below. do we really need this?

boxblinkracer (commented):
please see comments

Subscriber/OrderBackendSubscriber.php
StephanAltmann85: It is required to get the content of the comments column.

We could load the comments via Ajax. Therefore we need an additional controller action and - because we should not load them one by one - some additional logic in ExtJs.

Facades/Notifications/Notifications.php
boxblinkracer: this was indeed a bug,
and i also found another bug, that a non-mollie order history has been marked as mollie if the webhook is called again.
so there was no check, if a new entry was even made by mollie.

i havent heard back from you in this PR, so i fixed it myself and will decline this one

PR #141, opened 2021-02-10
MOL-158: Test if getInvoiceShippingTaxRate Method exists

ThLind:
In earlier versions of Shopware the getInvoiceShippingTaxRate Method as well as the invoice_shipping_tax_rate column in the order table don't exist. So check if the method exists, otherwise calculate the tax rate by invoice shipping and invoice shipping net.

### DIFF
--- a/Components/Shipping/Providers/CartShippingCostsProvider.php
+++ b/Components/Shipping/Providers/CartShippingCostsProvider.php
@@ -86,7 +86,13 @@ class CartShippingCostsProvider implements ShippingCostsProviderInterface
      */
     private function getTaxRate(Order $order)
     {
-        $taxRate = $order->getInvoiceShippingTaxRate();
+        $taxRate = null;
+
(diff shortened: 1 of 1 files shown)


Components/Shipping/Providers/CartShippingCostsProvider.php
```diff
@@ -86,7 +86,11 @@ private function getInvoiceShippingCosts()
      */
     private function getTaxRate(Order $order)
     {
-        $taxRate = $order->getInvoiceShippingTaxRate();
+        $taxRate = null;
+
+        if (method_exists($order, 'getInvoiceShippingTaxRate')) {
+            $taxRate = $order->getInvoiceShippingTaxRate();
```
boxblinkracer, line 92: what about DHL wunschpaket in that scenario?

ThLind: This doesn't affect DHL Wunschpaket.

PR #142, opened 2021-02-11
MOL-53: Apple Pay Button after offcanvas change

Sironheart:


barbieswimcrew (approved):

Resources/views/frontend/checkout/ajax_cart.tpl
```diff
@@ -7,6 +7,16 @@
             <div class="apple-pay--container apple-pay--container--ajax-cart">
                 {include 'frontend/plugins/payment/mollie_applepay_direct.tpl' }
             </div>
+            <script type="text/javascript">
+                // This needs to be inline to force the script on offcanvas reload
+                var applePayContainer = $('.apple-pay--container--ajax-cart');
```
boxblinkracer, line 12: can we move this to javascript files please?

boxblinkracer (changes requested):

Resources/views/frontend/checkout/ajax_cart.tpl
Sironheart: Sadly not, since there seems to be some issue with safari.

boxblinkracer: ? all apple pay javascript code is placed in js files? that makes no sense?

Sironheart: as discussed, we need to do the initApplePay functionality inline.

boxblinkracer (approved):

boxblinkracer:
@Sironheart  can you fix the conflicts? then i would merge it, thanks

PR #149, opened 2021-02-22
MOL-247: Ensure Payment is set in Session

ThLind:
When the payment isn't set in session restore it from transaction before creating order.

### DIFF
--- a/Exceptions/PaymentMeanNotFoundException.php
+++ b/Exceptions/PaymentMeanNotFoundException.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+
+namespace MollieShopware\Exceptions;
+
+
+use Exception;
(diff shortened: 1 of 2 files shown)


barbieswimcrew (approved):

boxblinkracer:
hi
i have a problem with this PR
...i even did a PR in the past to automatically restore the payment ID...

why do we think only the payment ID is missing? because its the first error obviously...but that does not make sense at all.
so i think the whole session is gone...to what would that lead? there will not be an error on the confirm page because nothing exists at all? and the previous order cannot even be restored or cancelled too.

so basic question: what happens? why is the paymentID null?

boxblinkracer (changes requested):
im quite sure this does not solve the problem

PR #153, opened 2021-03-01
Fix Missing Expired Mapping in Order API notifications

boxblinkracer:
expired status mapping was just missing for the orders api
that means there was an error in the notification when expired has been received

i did also improve the logs a bit

### DIFF
--- a/Components/Helpers/MollieStatusConverter.php
+++ b/Components/Helpers/MollieStatusConverter.php
@@ -70,6 +70,10 @@ class MollieStatusConverter
                 $targetStatus = PaymentStatus::MOLLIE_PAYMENT_FAILED;
             }
 
+            if ($paymentsResult[PaymentStatus::MOLLIE_PAYMENT_EXPIRED] == $paymentsResult['total']) {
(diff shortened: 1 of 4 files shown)


Components/Helpers/MollieStatusConverter.php
```diff
@@ -70,6 +70,10 @@ public function getMollieOrderStatus(Order $order)
                 $targetStatus = PaymentStatus::MOLLIE_PAYMENT_FAILED;
             }
 
+            if ($paymentsResult[PaymentStatus::MOLLIE_PAYMENT_EXPIRED] == $paymentsResult['total']) {
```
barbieswimcrew, line 73: might be a typesafe condition here?

barbieswimcrew (changes requested):

Components/Helpers/MollieStatusConverter.php
boxblinkracer: na please not, not going to touch that code for now, will be changed anyway soon
but typesafe already brought some problems in that code *sigh :D

schmitzcarsten (approved):

PR #155, opened 2021-03-09
MOL-197: Improved getting Mollie Orders and Payments

ThLind:
In the Logs of MOL-197 can be found the following Error: "error":"Call to a member function getMollieOrderId() on null"
This can only happen in PaymentService, as at all other places where this method is called it's checked before the method call if the object is of the right type.
In PaymentService I found unnecessary loading of the transaction to get the orderId (which caused the error when the loading returned null) although the transaction is already present and checked in all places where this method of the PaymentService is used, so I added the orderId as parameter for the method and don't try to load the transaction. 
The exactly same happend in the method to load the Mollie Payment.


barbieswimcrew (approved):

boxblinkracer (changes requested):
please resolve conflicts

PR #157, opened 2021-03-15
MOL-98: ApplePay payment methods not as default payment

Sironheart:


Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -27,9 +63,82 @@ public static function getSubscribedEvents()
             'Enlight_Controller_Action_PostDispatch_Frontend' => 'onFrontendPostDispatch',
             'Enlight_Controller_Action_PostDispatch_Widget' => 'onFrontendPostDispatch',
             'Enlight_Controller_Action_PostDispatch_Frontend_Checkout' => 'onFrontendCheckoutPostDispatch',
+            'Enlight_Controller_Action_PreDispatch_Frontend_Checkout' => 'writeDefaultPaymentToSession',
```
boxblinkracer, line 66: please name functions with the event that it does. so we know what is happening
in that case "onFrontendCheckoutPreDispatch"

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -27,9 +63,82 @@ public static function getSubscribedEvents()
...
+            'Shopware_Modules_Admin_UpdatePayment_FilterSql' => 'onApplePayButtonChangePaymentMethod',
         ];
     }
 
+    public function writeDefaultPaymentToSession(\Enlight_Event_EventArgs $args)
+    {
+        if (
```
boxblinkracer, line 73: is this something like 
if ($isDefaultPaymentAllowed) ?
then please class + unit test, thx

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -27,9 +63,82 @@ public static function getSubscribedEvents()
...
+
+        if ($request->getActionName() !== 'finish') {
+            return;
+        }
+
+        $userId = $this->adminModule->sGetUserData()['additional']['user']['id'];
+
+        $customer = $this->modelManager->find(Customer::class, $userId);
+
+        $sql = $this->modelManager->getConnection()->prepare(
```
boxblinkracer, line 94: please no sql in here, encapsulate it to the corresponding tier, thx

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -27,9 +63,82 @@ public static function getSubscribedEvents()
...
+            $paymentId = $result['paymentID'];
+        }
+
+        $customer->setPaymentId($paymentId);
+
+        $this->modelManager->persist($customer);
+        $this->modelManager->flush();
+    }
+
+    public function onApplePayButtonChangePaymentMethod(\Enlight_Event_EventArgs $args)
```
boxblinkracer, line 120: as described above, regarding to the event, this is more like the onUpdatePaymentMethod?

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -27,9 +63,82 @@ public static function getSubscribedEvents()
...
+    {
+        $subject = $args->get('subject');
+        $returnValue = $args->getReturn();
+        $paymentMeans = $subject->sGetPaymentMeans();
+
+        $applePayPayments = \array_filter($paymentMeans, static function($paymentMean) {
```
boxblinkracer, line 126: i read, but somehow dont get the purpose within an appropriate time of thinking.
it does change to ANY apple pay payment?
looks too complicated, has sql, ...
please improve wording, paragraphs and encapsulate what is necessary, thx

boxblinkracer (changes requested):
please see comments

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Restructured. Unit test not added yet.

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: I've rebuild it for reducing cognitive load.

It's a Filter_SQL_Event - we do replace the Update-Query with a Select-Query to prevent updating the user's default payment method.

boxblinkracer:
hi, I've tested it before reviewing, 
didnt work
video will be added to jira ticket

StephanAltmann85:
> hi, I've tested it before reviewing,
> didnt work
> video will be added to jira ticket

My fault. Fixed and tested

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -14,10 +19,41 @@
...
+use Shopware\Components\Model\ModelManager;
 use Shopware\Components\Theme\LessDefinition;
+use Shopware\Models\Customer\Customer;
+use Shopware\Models\Order\Order;
+use Shopware_Components_Config;
+use Shopware_Components_Modules;
 
 class ApplePayDirectSubscriber implements SubscriberInterface
 {
+    const APPLEPAY_NAME = 'mollie_applepay';
```
boxblinkracer, line 31: please use constants of class "ShopwarePaymentMethod".
thats already existing

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -115,6 +194,55 @@ public function onFrontendCheckoutPostDispatch(Enlight_Event_EventArgs $args)
         $view->assign('sPayments', $sPayments);
     }
 
+    /**
+     * @param int $customerId
+     * @param int $paymentId
+     * @throws \Doctrine\ORM\ORMException
+     * @throws \Doctrine\ORM\OptimisticLockException
+     */
+    private function updateCustomerDefaultPaymentMethod(int $customerId, int $paymentId) {
```
boxblinkracer, line 203: could you auto-format that file
the { and a few other things tells me it has not been done, just a small thing, thank you

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -115,6 +194,55 @@ public function onFrontendCheckoutPostDispatch(Enlight_Event_EventArgs $args)
...
+        $this->modelManager->flush();
+    }
+
+    /**
+     * @param int $customerId
+     * @return mixed|null
+     * @throws \Doctrine\ORM\NoResultException
+     * @throws \Doctrine\ORM\NonUniqueResultException
+     */
+    private function getCustomerDefaultNonApplePayPaymentMethod(int $customerId) {
```
boxblinkracer, line 219: please do not use sql or queries in a subscriber
please place it in a service / component..just anywhere else
theres an Account componentn, that would be a great thing i guess...to get xy from the current account
that also helps us to do some refactoring later on when restructuring all these services

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -115,6 +194,55 @@ public function onFrontendCheckoutPostDispatch(Enlight_Event_EventArgs $args)
...
+        $paymentId = $this->config->get('defaultpayment');
+
+        $queryBuilder = $this->modelManager->createQueryBuilder();
+        $queryBuilder
+            ->select('orders.paymentId')
+            ->from(Order::class, 'orders')
+            ->leftJoin('orders.payment', 'payment')
+            ->where('orders.customerId = :customerId')
+            ->andWhere('payment.name != :applePayName')
```
boxblinkracer, line 228: please always use expressions $expr->notEqual..... thank you

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -115,6 +194,55 @@ public function onFrontendCheckoutPostDispatch(Enlight_Event_EventArgs $args)
...
+            ->andWhere('payment.name != :applePayDirectName')
+            ->orderBy('orders.id', 'DESC')
+            ->setParameter('customerId', $customerId)
+            ->setParameter('applePayName', self::APPLEPAY_NAME)
+            ->setParameter('applePayDirectName', self::APPLEPAYDIRECT_NAME)
+            ->setMaxResults(1)
+        ;
+
+        $customerOrders = $queryBuilder->getQuery()->getSingleResult();
```
boxblinkracer, line 237: what happens if the customer has no order? doesnt getSingleResult imply a single result?
so i guess there is an exception and the if from below isnt triggered in that case?

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -170,4 +298,18 @@ private function preventApplePayDirectAsPaymentMethodOnConfirm(Enlight_View $vie
         $view->assign('sUserData', $userData);
     }
 
+    private function isDefaultPaymentMethodApplePay() {
```
boxblinkracer, line 301: i would need a unit test for this please, so that its documented and verified for the future

thus please move to a separate class or existing component, and adjust the payment name as argument
then create a few tests to verify the correctly configured apple pay "restrictions" for this use case

thank you

boxblinkracer (changes requested):
please see comments

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Fixed

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: csfixer implementation would solve this

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: static code analysis would prevent those issues

Subscriber/ApplePayDirectSubscriber.php
boxblinkracer: thats the reason why i'm currently adding it anyway in the meanwhile

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Unit test for method added

StephanAltmann85:
> please see comments

done

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -3,21 +3,60 @@
...
+
+    /**
+     * @var ApplePayPaymentMethod
+     */
+    private $paymentMethodService;
+
+    /**
+     * @param Shopware_Components_Modules|null $modules
+     */
+    public function __construct(Account $accountService, ApplePayPaymentMethod $paymentMethodService, $modules = null)
```
boxblinkracer, line 53: hi
could you try either try it with this

` <service id="modules" synthetic="true" />`

or just acessing the Shopware()->Modules in the constructor or where existing...
but we would need to get rid off the nullable object in the contructor
        
thank you

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -26,9 +65,62 @@ public static function getSubscribedEvents()
...
 
+    /**
+     * @param Enlight_Event_EventArgs $args
+     */
+    public function onFrontendAccountPostDispatch(\Enlight_Event_EventArgs $args)
+    {
+        /** @var Enlight_Controller_Request_RequestHttp $request */
+        $request = $args->get('request');
+
+        if ($request->getActionName() !== 'login') {
```
boxblinkracer, line 81: i totally forgot to test the account in the first run.
i did now, but its not working
probably due to this line?

from what i see, it should prevent the user from saving apple pay as default method, but 2 things happen

a) this code isnt triggered at all, because it returns here

b) apple pay and direct are accidentally visible here. we would need to (probably) hide it in chrome like on the checkout payment selection page...but this is also something I can do afterwards...but please let me know
also apple pay direct should always be hidden (but thats also out of this scope which is ok for me)

Subscriber/ApplePayDirectSubscriber.php
```diff
@@ -26,9 +65,62 @@ public static function getSubscribedEvents()
...
+    public function onFrontendCheckoutPreDispatch(\Enlight_Event_EventArgs $args)
+    {
+        /** @var Enlight_Controller_Request_RequestHttp $request */
+        $request = $args->get('request');
+
+        if ($request->getActionName() !== 'finish') {
+            return;
+        }
+
+        $currentPaymentMethod = $this->modules->Admin()->sGetUserData()['additional']['payment']['name'];
```
boxblinkracer, line 113: just something that popped up in my mind
if this code breaks, will the checkout also break?
if so we would need a try/catch and a logger with a warning...because that feature must never break a checkout

Tests/Components/ApplePayDirect/Services/ApplePayPaymentMethodTest.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    private $paymentMethodService;
+
+    public function setUp(): void
+    {
+        $this->paymentMethodService = new ApplePayPaymentMethod(
+            $this->createMock(PaymentMethodService::class)
+        );
+
+        require_once __DIR__ . '/../../../Shopware/Components/Plugin.php';
```
boxblinkracer, line 19: why do we need this in here? shopware is a dev dependency?

Tests/Components/ApplePayDirect/Services/ApplePayPaymentMethodTest.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Tests if isApplePayPaymentMethod will return false if given method name is not name of Apple Pay method
+     *
+     * @covers \MollieShopware\Components\ApplePayDirect\Services\ApplePayPaymentMethod
+     */
+    public function testIsApplePaymentMethodWillReturnFalseOnOtherMethod()
+    {
+        $this->assertFalse($this->paymentMethodService->isApplePayPaymentMethod('mollie_creditcard'));
```
boxblinkracer, line 29: could you use data providers please and do only 1 assertion - thank you

Tests/Shopware/Components/Plugin.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Shopware\Components;
```
boxblinkracer, line 3: i think we dont need this because we have shopware as dev dependency right?

boxblinkracer (changes requested):
thanks for the changes. way better now
but i found a few things

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: This will always be set when calling sGetuserData:

`            $userData['additional']['payment'] = $this->sGetPaymentMeanById(
                $userData['additional']['user']['paymentID'],
                $userData
            );`

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Well, this is just a fallback if you have selected ApplePay as payment method but have not finished the order. On login while switching the browser the default payment method will be reseted.

On the account page for selecting default payment method apple pay could be removed from payment means as done with Apple Pay Direct.

Subscriber/ApplePayDirectSubscriber.php
boxblinkracer: i was talking about this whole code and process...so if something breaks when updating the payment method

Tests/Shopware/Components/Plugin.php
StephanAltmann85: The constants in ShopwarePaymentMethod do refer to MollieShopware. This class does extend from Plugin. The kernel in the unit tests is not loaded correctly. 

I haven't checked the pipeline, but when using the shopware release instead of the repository the '\Tests\bootstrap.php' is missing. For just having this one class the whole autoloading in the unit tests has to be changed.

Tests/Components/ApplePayDirect/Services/ApplePayPaymentMethodTest.php
StephanAltmann85: The constants in ShopwarePaymentMethod do refer to MollieShopware. This class does extend from Plugin. The kernel in the unit tests is not loaded correctly.

I haven't checked the pipeline, but when using the shopware release instead of the repository the '\Tests\bootstrap.php' is missing. For just having this one class the whole autoloading in the unit tests has to be changed.

Subscriber/ApplePayDirectSubscriber.php
boxblinkracer: so for the test cases:
i use SOFORT and complete the order,
then i start using apple pay in safari, but do not place the order...then i switch to chrome...login..and there it sets it to SOFORT and is immediately visible on the confirm page?

does it really need to happen in the login? cant we move it to somewhere else?
like in account and checkout in those spots?

also yes please remove apple pay direct, and hide apple pay on account page, thank you

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: I've added try catch just to be sure

Tests/Shopware/Components/Plugin.php
boxblinkracer: but it works for me?

shopware-5_–_ShopwarePaymentMethodTest_php

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: If we do it in checkout the payment method would change without interaction during the process. Except while logging account is maybe never accessed.

Tests/Components/ApplePayDirect/Services/ApplePayPaymentMethodTest.php
boxblinkracer: works for me in other screenshot

Subscriber/ApplePayDirectSubscriber.php
boxblinkracer: thank you

Subscriber/ApplePayDirectSubscriber.php
boxblinkracer: thank you

Tests/Shopware/Components/Plugin.php
StephanAltmann85: I've rebased after test implementation. Before it was not working. Yes, wo don't need it anymore.

boxblinkracer: perfect, thx, the dev dependency has been added a while ago

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: Now hidden on account page

Tests/Components/ApplePayDirect/Services/ApplePayPaymentMethodTest.php
StephanAltmann85: Done

Subscriber/ApplePayDirectSubscriber.php
StephanAltmann85: I've simply moved the methods to another subscriber because they are not only ApplePayDirect related. So we won't need that nullable arg anymore.

Subscriber/ApplePaySubscriber.php
```diff
@@ -0,0 +1,129 @@
...
+        try {
+            if ($this->paymentMethodService->isApplePayPaymentMethod($currentPaymentMethod) === false) {
+                return;
+            }
+
+            $userId = (int)$this->modules->Admin()->sGetUserData()['additional']['user']['id'];
+            $paymentId = $this->accountService->getCustomerDefaultNonApplePayPaymentMethod($userId);
+
+            $this->accountService->updateCustomerDefaultPaymentMethod($userId, $paymentId);
+        } catch (Throwable $e) {
```
boxblinkracer, line 84: achtung php 5.6 hat keine throwable
bitte exception machen und eine kleine warning log (das hätte ich glaube ich mal gewünscht)

Subscriber/ApplePaySubscriber.php
```diff
@@ -0,0 +1,129 @@
...
```
boxblinkracer, line 112: achtung php5.6, siehe oben...+ log bitte

boxblinkracer (changes requested):

Subscriber/ApplePaySubscriber.php
StephanAltmann85: Done

Subscriber/ApplePaySubscriber.php
StephanAltmann85: Done

boxblinkracer (approved):

PR #160, opened 2021-03-17
MOL-352: Basket Restore for SWAG Custom Products Plugin

ThLind:
Added Compatibility for Restore Basket (create order before Payment) for SWAG Custom Products Plugin

The SWAG Custom Products Plugin generates Hashes for the possible Configurations. These are stored in the OrderItem and read from the Request when the Item is added to the Basket. Depending on the Hash and the Configuration for the Hash it adds the configuration Items (Mode 4, Surcharge/Discount Mode) and the normal Execution of the sAddArticle Method from sBasket is stopped afterwards.

So for adding Compatibility for this Plugin check if the Hash exists in the orderDetails Object Attributes and if yes store it in the Request.

### DIFF
--- a/Components/Services/BasketService.php
+++ b/Components/Services/BasketService.php
@@ -8,6 +8,7 @@ use MollieShopware\Components\Config;
 use MollieShopware\Components\Logger;
 use MollieShopware\Components\TransactionBuilder\Models\BasketItem;
 use Psr\Log\LoggerInterface;
+use Shopware\Components\DependencyInjection\Container as DIContainer;
 use Shopware\Components\Model\ModelManager;
 use Shopware\Models\Order\Basket;
 use Shopware\Models\Order\Detail;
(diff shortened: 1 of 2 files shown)


barbieswimcrew (approved):

Components/Services/BasketService.php
```diff
@@ -40,11 +42,16 @@ class BasketService
...
+    private $container;
+
     /**
      * @param ModelManager $modelManager
      * @param LoggerInterface $logger
+     * @param DIContainer $container
+     * @throws Exception
      */
-    public function __construct(ModelManager $modelManager, LoggerInterface $logger)
+    public function __construct(ModelManager $modelManager, LoggerInterface $logger, DIContainer $container)
```
boxblinkracer, line 54: i know this is werid, but can you remove the type "DIContainer"
i added that once and it failed in a shop within shopware...it was another incompatible class that was used
you can keep it in line 45 as annotation, but just remove it in the constructor, thx

boxblinkracer (changes requested):
please see comments

Components/Services/BasketService.php
ThLind: Ok, removed it.

PR #171, opened 2021-03-25
NTR: add PHP CS Fixer

boxblinkracer:


boxblinkracer:
add cs fixer
with 1x CI workflow check in dry run

+ make PR command including a fix

also added the modified files

schmitzcarsten (approved):

PR #178, opened 2021-04-06
MOL-247: Fix Missing Orders

boxblinkracer:
do not merge, thats a test for now


boxblinkracer:
added session token to every transaction.
if a return action is happening without a session, it will do different checks including token comparing to verify if its allowed to restore the session.

if its allowed, we set the previous shopware cookie, which automatically restores the correct session data on the next request.
from that point on, everything is working as expected.

I've added a SessionManager class to create tokens, delete tokens and restore a session from a token.

a token is always deleted immediately as well as in webhooks if the user is not coming to the shop (expiration)

i've added new Cypress tests that prove the loss of a session and also added new plugin configuration changes using the shopware backend to really test create-order-BEFORE-payment and create-order-AFTER-payment in Shopware


besides that a payment status resolver has been created to fetch the mollie status of a transaction.
this is just to make webhooks easier and cleaner

schmitzcarsten (approved):

PR #192, opened 2021-04-23
MOL-172: Test API in plugin

boxblinkracer:
add test api button that iterates through all shops and tests their api keys

result is in extjs notification as well as in logs


boxblinkracer:
please note, this pr might fail at the moment (pipeline)
trying to debug if pipeline finds all stuff :)

barbieswimcrew (approved):

PR #195, opened 2021-04-28
MOL-370: Refactor payment method requests

boxblinkracer:


boxblinkracer:
i know these are really a lot of files
did private code review session with MW due to the load of files

the PR creates separate request objects for all payment methods, so payment method object you can really use and see how they would look like

including refactorings, unit tests for the json requests and way more
unit tests + cypress tests pass

Tests/PHPUnit/Services/Mollie/Payments/Converters/AddressConverterTest.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace MollieShopware\Tests\Services\Mollie\Payments\Builders;
+
+use MollieShopware\Services\Mollie\Payments\Converters\AddressConverter;
+use MollieShopware\Services\Mollie\Payments\Models\PaymentAddress;
+use PHPUnit\Framework\TestCase;
+
+class AddressConverterTest extends TestCase
```
barbieswimcrew, line 9: @boxblinkracer you're sure you want to use the actual address data instead of a fake one?

barbieswimcrew (changes requested):

Tests/PHPUnit/Services/Mollie/Payments/Converters/AddressConverterTest.php
boxblinkracer: yes, because this class is about testing that AddressConverter :)

boxblinkracer: ok i have no clue what you mean.
i test the payment-to-array converter by providing a fixture payment and assert the resulting array?! :)

barbieswimcrew (approved):

PR #208, opened 2021-05-21
MOL-182: Add orders expiration days

boxblinkracer:


boxblinkracer:
this feature brings a new expiration orders config to every payment method.

thus it adds a new feature, to add custom mollie configs directly in shopware extJS backend in the payment methods window.
there is a new section now for mollie configs such as the expirationDays one.
this can also be translated for sub shops.

saving is done in a new configuration entity and not as attributes fields.

and when a payment is started, it will ask for that data if existing or if translated and just use it
in the payment request body

barbieswimcrew (approved):

schmitzcarsten (approved):

PR #210, opened 2021-05-25
MOL-105: add tracking number to shipping

boxblinkracer:


boxblinkracer:
if shopware has tracking information, then we send this to mollie too.
i've created a central component that reads the correct data from the shopware order and passes that optional data on to Mollie

Components/Mollie/MollieShipping.php
```diff
@@ -0,0 +1,55 @@
...
+use Shopware\Models\Dispatch\Dispatch;
+use Shopware\Models\Order\Order;
+
+class MollieShipping
+{
+
+    /**
+     * Shopware tracking code variable of order.
+     */
+    const TRACKING_CODE_VARIABLE = '{$sOrder.trackingcode}';
```
uehler, line 15: I don't know if it's relevant in this case but there is also $offerPosition.trackingcode: https://docs.shopware.com/de/shopware-5-de/tutorials-und-faq/trackingnummern-und-sendungsverfolgung

boxblinkracer: we only support the overall order tracking number for now :)

PR #211, opened 2021-05-26
MOL-181: Payment Link in Shopware Backend

boxblinkracer:
added new live preview of the mollie payment data in the shopware backend.

the orders detail overview has been extended with a new section
that shows data for the payment

the data will be fetched from a backend controller.

there are also 2 copy clipboard buttons and a deep-link button to open the payment in the mollie dashboard

### DIFF
--- a/Models/TransactionRepository.php
+++ b/Models/TransactionRepository.php
@@ -25,16 +25,14 @@ class TransactionRepository extends ModelRepository implements TransactionReposi
     }
 
     /**
-     * Get the most recent transaction for an order
-     *
-     * @param Order $order
-     * @return Transaction|null
+     * @param int $orderId
+     * @return Transaction
      */
-    public function getMostRecentTransactionForOrder(Order $order)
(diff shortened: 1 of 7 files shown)


Models/TransactionRepository.php
```diff
@@ -25,16 +25,14 @@ public function save(Transaction $transaction)
     }
 
     /**
-     * Get the most recent transaction for an order
-     *
-     * @param Order $order
-     * @return Transaction|null
+     * @param $orderId
```
uehler, line 28: type hint

Models/TransactionRepository.php
```diff
@@ -25,16 +25,14 @@ public function save(Transaction $transaction)
...
+     * @return Transaction
      */
-    public function getMostRecentTransactionForOrder(Order $order)
+    public function getTransactionByOrder($orderId)
```
uehler, line 31: [ ] type hint

PR #217, opened 2021-06-14
MOL-489: Skip Klarna Paid Status Mails

boxblinkracer:
if payment status mails are enabled, the customer gets a paid-mail when mollie sends "paid" or completed for klarna.

the problem is, that this is only between the merchant and mollie.

the customer himself did not pay to klarna yet.

this mail would confuse him, so we make sure to always skip it

### DIFF
--- a/Components/Order/OrderUpdater.php
+++ b/Components/Order/OrderUpdater.php
@@ -6,9 +6,9 @@ use Doctrine\ORM\EntityRepository;
 use Doctrine\ORM\OptimisticLockException;
 use Doctrine\ORM\ORMException;
 use MollieShopware\Components\Config;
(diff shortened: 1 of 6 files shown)


barbieswimcrew (approved):

Components/Order/OrderUpdater.php
```diff
@@ -53,6 +52,11 @@ class OrderUpdater
      */
     private $repoHistory;
 
+    /**
+     * @var PaymentStatusMailValidator
+     */
+    private $validatorPaymentMail;
```
schmitzcarsten, line 58: just a naming, but isn't it better to name variable like class. The variable name sounds strange

schmitzcarsten (approved):

PR #220, opened 2021-06-14
MOL-449: Support Shopware 5.7.0

boxblinkracer:


boxblinkracer:
had to set some services to public in xml
and also change the logger
its not inherited anymore, but wraps a logger internally

the introspection still seems to work fine

also had to adjust pipelines and dev-dependencies to be based on php 7.4 in order to work with the latest updates

schmitzcarsten (approved):

PR #223, opened 2021-06-15
MOL-443: Add Apple Pay Direct Risk Management Support

boxblinkracer:


boxblinkracer:
added support for risk management

this will be done directly in the subscriber when the visibility of the button is determined.
or directly when finishing the apple pay payment. this is necessary for risk rules with user data...errors will lead to seeing the confirrm page with a custom "Risk Management" error text

Resources/snippets/frontend/mollie/plugins.ini
```diff
@@ -37,6 +38,7 @@ CardNumberLabel="Kreditkartennummer"
 CardExpiryDateLabel="Ablaufdatum"
 CardVerificationCodeLabel="CVC"
 CreditCardTagLine="Zahlung gesichert und bereitgestellt von"
+BlockedDueRiskManagement="Diese Zahlungsart ist leider für diesen Kauf nicht verfügbar. Sie können eine andere Zahlungsart und wählen und Ihren Einkauf fortführen."
```
schmitzcarsten, line 41: Zahlungsart und wählen

delete the und here

schmitzcarsten (changes requested):

Resources/snippets/frontend/mollie/plugins.ini
boxblinkracer: uuuu danke :D

schmitzcarsten (approved):

PR #225, opened 2021-06-17
WIP: MOL-493: Remove Klarna Shipped Status Configuration

boxblinkracer:
this function did only update the order status to shipped when doing a manual shipping in shopware.

it doesnt work in other shipping ways though

and also, if the update order status feature is activated, it will be overwritten anyway with the incoming webhook


boxblinkracer:
has to be discussed internally, but you can already do a review

schmitzcarsten (approved):

boxblinkracer:
there will be a new approach on that feature

PR #241, opened 2021-07-06
MOL-493: Improve Shipping Order Status

boxblinkracer:


boxblinkracer:
theres a new status "shipping" that only happens on partial shipments.
if the status is somehow "shipping", then we have a new behaviour
a) do not update payment status for shipping
b) update order status if configured for such a state

schmitzcarsten (approved):

PR #243, opened 2021-07-08
MOL-507: OrderCreation as payment configuration

boxblinkracer:


Resources/snippets/backend/mollie/general.ini
```diff
@@ -93,8 +102,16 @@ payment_details_method_readmore = "Mehr erfahren"
 payment_details_method_value_global = "Globale Einstellung (Plugin Konfiguration)"
 payment_details_method_value_paymentapi = "Payments API (Transaktionen)"
 payment_details_method_value_ordersapi = "Orders API (Transaktionen + Bestellungen)"
-payment_details_order_expires_caption = "Verfalldatum der Bestellung (Tage)"
+
+payment_details_ordercreation_caption = "Bestellung Erstellung"
```
boxblinkracer, line 106: this caption is still in discussion

boxblinkracer:
this PR brings the configuraiton "create order before or after payment" to the payments as payment specific config.

every payment method can either have
* GLOBAL SETTING
* CREATE ORDER BEFORE PAYMENT
* CREATE ORDER AFTER PAYMENT

please pay attention to codes like this
$orderCreation === OrderCreationType::BEFORE_PAYMENT

these will verify that the functionality stays the same

there is a new PaymentConfigResolver that easily gets the final config for such a payment configuration.
this depends on snippet translations, or the plain payment config without translation, or even the global PLUGIN configuration.
so its a nice helper


also the cypress tests have been adjusted and a bit restructured for the payment specific configs

Components/Config/PaymentConfigResolver.php
```diff
@@ -0,0 +1,196 @@
...
+        # for our provided shop
+        $translatedValue = $this->translations->getPaymentConfigTranslation(
+            ConfigurationKeys::METHODS_API,
+            $paymentConfig->getPaymentMeanId(),
+            $shopId
+        );
+
+        # use either the snippet or the
+        # value from the config directly
+        $methodType = (!empty($translatedValue)) ? (int)$translatedValue : (int)$paymentConfig->getMethodType();
```
schmitzcarsten, line 83: Why is it a int cast? Maybe better name methodTypeId or something like that? I would await string

schmitzcarsten, line 83: Why is it a int cast? Maybe better name methodTypeId or something like that? I would await string

schmitzcarsten (approved):

Components/Config/PaymentConfigResolver.php
boxblinkracer: its no realy foreign key ID, the type is an enum based on an INT db field
do you still think ID would be better?

PR #265, opened 2021-09-20
MOL-598: Migration Problem of Plugin Configuration

boxblinkracer:


boxblinkracer:
somehow weird values are existing in the database due to a migration, maybe empty string?
but this PR makes sure, that (in that case) the global plugin config will be used

barbieswimcrew (approved):

PR #269, opened 2021-09-27
MOL-578: Simple Bank Transfer Flow

boxblinkracer:


Subscriber/MailSubscriber.php
```diff
@@ -0,0 +1,248 @@
...
+     * them in the database to use them when the order is fully processed.
+     *
+     * @param Enlight_Event_EventArgs $args
+     * @return false|void
+     * @throws \Doctrine\ORM\ORMException
+     * @throws \Doctrine\ORM\OptimisticLockException
+     */
+    public function onSendMail(Enlight_Event_EventArgs $args)
+    {
+        $variables = $args->get('variables');
```
boxblinkracer, line 117: this is the old function only with some early returns
logic did not change

boxblinkracer:
Code Review Information:
there will be a new feature to redirect the user immediately to the shopware finish page instead of the external mollie page that might be in between. this feature makes the way "shorter".
its only for bank transfer

theres a new feature for bank transfer, either normal or easy flow option.

the checkout process will then use that logic and redirect to finish immediately (just like with Apple pay for instance).

the mail subscriber has been added,
the existing sendMail function was moved into that file and refactored a bit.
a new listener is added that adds variables for the bank data to be used in the order confirmation mail by the merchant.
the new easy way would not show bank data, so they have to be extracted dynamically from tthe payment

barbieswimcrew (approved):

Resources/views/backend/mollie_extend_payment/view/payment/form_panel.js
```diff
@@ -142,6 +146,43 @@ Ext.define('Shopware.apps.Mollie.Payment.view.payment.FormPanel', {
...
+                        },
+                        {
+                            xtype: 'button',
+                            flex: 2,
+                            text: me.molSnippets.methodReadMore,
+                            cls: 'small primary',
+                            margin: '2 0 0 10',
+                            scope: this,
+                            handler: function () {
+                                window.open('https://github.com/mollie/Shopware/wiki/SEPA-Bank-Transfer', '_blank').focus();
```
schmitzcarsten, line 181: Nice! Fetching Wiki from github. But if you change the wiki because of a newer version, the old versions would get the updated information from github. Is this really ok? Because it could be information that is not correct for the old version

schmitzcarsten (approved):

Resources/views/backend/mollie_extend_payment/view/payment/form_panel.js
boxblinkracer: this is no fetch, it should just open the wiki page with further instructions on the features. and this will be done in the bank transfer instructions page.
thats a requirement

PR #296, opened 2021-11-17
MOL-549: Fix Orders API calculation for VAT free countries

boxblinkracer:


SandyMental:
can i just add the one changed file to our installation and we are good to go for VAT free countries?

boxblinkracer:
hi @SandyMental 
yes, that should be fine
but please keep in mind, i'm currently working on it.

but i just did the last force push
so once the PR pipe is green, you could give it a test (and please give me feedback haha)

thanks

PR #300, opened 2021-11-26
MOL-145 import payment methods from all api keys

iamreinder:
Added an iteration over shops to get active payment methods for each subshop which might have different API keys.

### DIFF
--- a/Components/Installer/PaymentMethods/PaymentMethodsInstaller.php
+++ b/Components/Installer/PaymentMethods/PaymentMethodsInstaller.php
@@ -5,10 +5,7 @@ namespace MollieShopware\Components\Installer\PaymentMethods;
 
 use Enlight_Template_Manager;
 use Mollie\Api\Exceptions\ApiException;
-use Mollie\Api\MollieApiClient;
-use Mollie\Api\Resources\BaseCollection;
 use Mollie\Api\Resources\Method;
-use Mollie\Api\Resources\MethodCollection;
 use MollieShopware\Components\Config;
 use MollieShopware\Components\Constants\BankTransferFlow;
 use MollieShopware\Components\Constants\OrderCreationType;
@@ -51,11 +48,6 @@ class PaymentMethodsInstaller
     private $modelManager;
 
     /**
-     * @var MollieApiClient
-     */
-    private $mollieApiClient;
-
-    /**
      * @var PaymentInstaller
      */
     private $paymentInstaller;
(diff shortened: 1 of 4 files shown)


Components/Installer/PaymentMethods/PaymentMethodsInstaller.php
```diff
@@ -574,27 +570,60 @@ private function setPaymentActive(Payment $payment, $isActive)
...
+                $mollieApiClient = $mollieApiFactory->create($shop->getId());
+            } catch (ApiException $e) {
+                $this->logger->error(
+                    sprintf('Error when creating a Mollie API client for shop %s', $shop->getName()),
+                    [
+                        'error' => $e->getMessage(),
+                    ]
+                );
+
+                continue;
```
boxblinkracer, line 597: could you change to just 1 try catch within the foreach please
its a all or nothing, but still fault tolerant on shop level, thats a good one

Components/Installer/PaymentMethods/PaymentMethodsInstaller.php
```diff
@@ -574,27 +570,60 @@ private function setPaymentActive(Payment $payment, $isActive)
     /**
      * Returns a collection of active payment methods from the Mollie API.
      *
-     * @return null|BaseCollection|MethodCollection
+     * @return array
      */
     private function getActivePaymentMethodsFromMollie()
```
boxblinkracer, line 575: there is still a api client as private member in this class.
i think we need to get rid off it right?
also, it might be good to use the api factory instead, then we have the factory as private member and always the need of thinking about what api client we need

boxblinkracer (changes requested):

Components/Installer/PaymentMethods/PaymentMethodsInstaller.php
iamreinder: Will do!

Components/Installer/PaymentMethods/PaymentMethodsInstaller.php
iamreinder: I already use the API factory. I create the factory within the `getActivePaymentMethodsFromMollie` method which is the only place it is used. So no need to pass that in the constructor. I'll remove the api client.

boxblinkracer: i, alright,
the constructor was more like a nice to have to move it to the dependency, but its not mandatory
only removing the wrong api client :)
helemaal bedankt

boxblinkracer (approved):

PR #302, opened 2022-01-04
MOL-186: Refunds for orders with vouchers

boxblinkracer:


boxblinkracer:
in case of voucher methods, its not possible. to refund the whole amount but only the amount excl. voucher value.
unfortunately mollie does not calculate this
so i've added a different way that creates a partial refund for such refunds where the partial value is the complete value without the voucher amount (aka the complete remaining value)

schmitzcarsten (approved):

PR #305, opened 2022-01-05
MOL-566 available payment methods based on limits

iamreinder:


Components/Payment/Provider/ActivePaymentMethodsProvider.php
```diff
@@ -0,0 +1,80 @@
...
+                    $parameters['resource'] = 'orders';
+                }
+
+                if (!in_array('includeWallets', $parameters, true)) {
+                    $parameters['includeWallets'] = 'applepay';
+                }
+
+                $activeMethods = $mollieApiClient->methods->allActive($parameters);
+
+                foreach ($activeMethods->getArrayCopy() as $method) {
```
schmitzcarsten, line 57: foreach in foreach. Please extract second foreach to another function or class to decrease complexity of code

schmitzcarsten (changes requested):

Components/Payment/Provider/ActivePaymentMethodsProvider.php
iamreinder: Yeah, you're right. That's better.

Components/Payment/Provider/ActivePaymentMethodsProvider.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        $methods = [];
+
+        if (empty($shops)) {
+            return $methods;
+        }
+
+        foreach($shops as $shop) {
+            try {
+                $methods = $this->getActivePaymentMethodsForShop($shop, $methods, $parameters);
```
boxblinkracer, line 45: wouldnt you require a methods merge if it iterates over shops?
otherwise it would always have the ones from the last shop?

Components/Payment/Provider/ActivePaymentMethodsProvider.php
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * @param Shop $shop
+     * @param array $methods
+     * @param array $parameters
+     * @return array
+     * @throws ApiException
+     */
+    private function getActivePaymentMethodsForShop(Shop $shop, array $methods = [], array $parameters = [])
+    {
+        $mollieApiFactory = new MollieApiFactory($this->config, $this->logger);
```
boxblinkracer, line 68: can you please add the service factory from the DI to the constructor of this class?
reason is that we always have 1 single factory that can be used,
only in the installation a newable is required because its done before the DI stuff exists,...but that should already exist in such a way

Components/Payment/Provider/ActivePaymentMethodsProvider.php
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        if (!in_array('includeWallets', $parameters, true)) {
+            $parameters['includeWallets'] = 'applepay';
+        }
+
+        $activeMethods = $mollieApiClient->methods->allActive($parameters);
+
+        foreach ($activeMethods->getArrayCopy() as $method) {
+            $methodIsInArray = !empty(array_filter($methods, static function ($item) use ($method) {
```
boxblinkracer, line 82: why do we provide existing methods in here, if the scope of this function is to return the ones from mollie?

Resources/config.xml
```diff
@@ -695,6 +695,56 @@
...
+            <label lang="en">Use Mollie payment method limits</label>
+            <label lang="de">Verwenden Sie die Limits der Mollie-Zahlungsarten</label>
+            <label lang="nl-NL">Gebruik Mollie's betaalmethodelimieten</label>
+            <label lang="es-ES">Utilice los límites de los métodos de pago de Mollie</label>
+            <label lang="fr-FR">Utiliser les limites des modes de paiement Mollie</label>
+            <label lang="it-IT">Utilizza i limiti del metodo di pagamento Mollie</label>
+            <value>no</value>
+            <description lang="en">Automatically hides payment methods in the checkout based on the limits of Mollie. For instance, if you can only accept credit card payments up to 500 euros, the checkout for a cart of 600 euros will not display credit card as a payment method.</description>
+            <description lang="de">Blendet automatisch Zahlungsarten im Checkout basierend auf den Limits von Mollie aus. Wenn Sie beispielsweise nur Kreditkartenzahlungen bis zu 500 Euro akzeptieren können, wird an der Kasse für einen Warenkorb von 600 Euro keine Kreditkarte als Zahlungsarte angezeigt.</description>
```
boxblinkracer, line 708: 1 e too much -> als Zahlungsart angezeigt

Subscriber/PaymentMethodSubscriber.php
```diff
@@ -0,0 +1,164 @@
...
+        $value = isset($basket['AmountNumeric']) ? $basket['AmountNumeric'] : '';
+        $currency = isset($basket['sCurrencyName']) ? $basket['sCurrencyName'] : '';
+
+        if ($value === '' || $currency === '') {
+            return;
+        }
+
+        # get all active and available payment methods for a certain amount from Mollie
+        $availableMethods = $this->getActivePaymentMethodsProvider($config)->getActivePaymentMethodsFromMollie(
+            [
```
boxblinkracer, line 77: can we get rid off that loose usage here, one could simple screw up a string param in here without recognizing..
can you learn our payment provider to "getByCartAmount"...then we always have 1 single source of truth for our use cases
also for the number_format, please use the NumberFormatter class from the mollie requests, i think that also exits, thank you

Tests/PHPUnit/Subscriber/PaymentMethodSubscriberTest.php
```diff
@@ -0,0 +1,126 @@
...
+     * are removed from the checkout view.
+     *
+     * @testdox Method removeUnavailablePaymentMethods removes the expected payment methods from view.
+     */
+    public function testRemovesPaymentMethods()
+    {
+        $available = $this->setUpAvailableMethods(['ideal', 'paypal']);
+
+        $paymentMethodSubscriber = new PaymentMethodSubscriber(
+            $this->createMock(LoggerInterface::class),
```
boxblinkracer, line 98: new NullLogger() :)

Tests/PHPUnit/Subscriber/PaymentMethodSubscriberTest.php
```diff
@@ -0,0 +1,126 @@
...
+            $this->createMock(MollieShopSwitcher::class)
+        );
+
+        $view = $this->setUpView();
+
+        $paymentMethodSubscriber->removeUnavailablePaymentMethods(
```
boxblinkracer, line 104: attention, this is not really testing what this test case implies.
i would think that this test verifies that our subscriber correctly removes the payment methods from the view.
but this actually just tests the "sub component" to remove methods according to a list of available payment methods.
if i would just remove the calling line in the subscriber itself, the test is still green, while the storefront/feature is broken.

so either you re-model it to test only the removing functionality, or you calll the "onFrontendCheckoutPostDispatch" function which might be the better one.
then you can also change the function to private inside the subscriber

boxblinkracer (changes requested):
please see comments

Subscriber/PaymentMethodSubscriber.php
boxblinkracer: sry for being confusing on this :D
i meant that the "free array" with configs shouldnt be passed in from here.
so that would then be done inside a getByCartAmount function.
otherwise one could use that getActivePaymentMethods in ANY way, which should not happen (at the moment at least) :)

PR #312, opened 2022-01-19
MOL-688: Apple Pay Direct support for required phone numbers

boxblinkracer:
this pull request makes sure to also require a phone number in apple pay direct, if configured in the shop.

the shop configuration will be passed on to apple pay, which will then add that requirement in javascript.
the data is then added in the post variables.

the php action will then also add the phone number to the registration.

the error handling has been improved in case something (or other keys) are not passed on.
violation exceptions now lead to a custom RegistrationInvalidField exception with more data about that field.

errors like these will then be shown on the PDP, because registration was not complete, means that the user cannot be forwarded to the cart to finish the checkout.
to show these errors on the PDP, an existing subscriber had to be modified to allow that, as well as a new smarty view to show custom errors on PDP.
instead of just using snippets in the frontend, i moved over to translate in PHP, which allows use to add that field to the template message. its also easier than the other approach

### DIFF
--- a/Traits/Controllers/RedirectTrait.php
+++ b/Traits/Controllers/RedirectTrait.php
@@ -61,6 +61,25 @@ trait RedirectTrait
 
     /**
      * @param \Enlight_Controller_Action $controller
+     * @param int $articleID
+     * @param string $errorMessage
+     * @throws \Exception
+     */
+    protected function redirectToPDPWithError(\Enlight_Controller_Action $controller, $articleID, $errorMessage)
+    {
+        Shopware()->Session()->offsetSet('mollieErrorMessage', $errorMessage);
+
+        $url = $controller->Front()->Router()->assemble([
+            'controller' => 'detail',
+            'action' => 'index',
+            'sArticle' => $articleID,
+        ]);
+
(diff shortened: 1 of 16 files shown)


Traits/Controllers/RedirectTrait.php
```diff
@@ -59,6 +59,25 @@ protected function redirectToShopwareCheckoutFailedWithError(\Enlight_Controller
...
+     * @param \Enlight_Controller_Action $controller
+     * @param int $articleID
+     * @param string $errorMessage
+     * @throws \Exception
+     */
+    protected function redirectToPDPWithError(\Enlight_Controller_Action $controller, $articleID, $errorMessage)
+    {
+        Shopware()->Session()->offsetSet('mollieErrorMessage', $errorMessage);
+
+        $url = Shopware()->Front()->Router()->assemble([
```
uehler, line 72: $controller->Front()->Router()->assemble

uehler (approved):

Traits/Controllers/RedirectTrait.php
boxblinkracer: thanks :)

PR #324, opened 2022-05-10
MOL-824 Mollie actions missing in order detail positions

iamreinder:


Facades/FinishCheckout/Services/ShopwareOrderUpdater.php
```diff
@@ -130,10 +128,9 @@ public function updateReferencesFromMollieOrder(Order $swOrder, \Mollie\Api\Reso
                     }
 
                     if (
-                        method_exists($detail->getAttribute(), 'getBasketItemId')
-                        && method_exists($detail->getAttribute(), 'setMollieTransactionId')
+                        method_exists($detail->getAttribute(), 'setMollieTransactionId')
                         && method_exists($detail->getAttribute(), 'setMollieOrderLineId')
-                        && (int)$detail->getAttribute()->getBasketItemId() === $transactionItem->getBasketItemId()
```
boxblinkracer, line 136: i think i recently heard that problem from a merchant who updated systems.
could we create an OR condition, so that either the NEW or the OLD way lead to to this? just to be safe,
rest looks totally fine :) thanks

boxblinkracer (changes requested):
see comments

Facades/FinishCheckout/Services/ShopwareOrderUpdater.php
iamreinder: It leads to weird behavior in my test shop. It always picks the shipping item. While matching SKUs works flawlessy in my testing.

PR #340, opened 2022-07-28
Feature/add subject to support form

iamreinder:


boxblinkracer:
Hi

why is there a node modules folder lock file? and also composer updates?
i dont see any reason why this should be in there?

PR #343, opened 2022-08-04
MOL-562 fixed display of Apple Pay payment method in OnePageCheckout

iamreinder:


Resources/views/frontend/_public/src/js/applepay.js
```diff
@@ -16,8 +16,17 @@
      */
     function hideApplePayIfNotAllowed() {
         // Find the hidden Apple Pay element
-        var applePayInput = document.getElementsByClassName('payment-mean-mollie-applepay');
-        var applePayLabel = document.getElementsByClassName('payment-mean-mollie-applepay-label');
+        var applePayInput = document.querySelector('input.payment-mean-mollie-applepay');
+        var applePayLabel = document.querySelector('label.payment-mean-mollie-applepay-label');
+
+        if (typeof applePayInput === 'undefined' || !applePayLabel) {
```
boxblinkracer, line 22: could you add comments what this actually does as a fallback detection?
i doubt anyone will remember this here otherwise

boxblinkracer (changes requested):
please see comments

Resources/views/frontend/checkout/change_payment.tpl
```diff
@@ -1,5 +1,15 @@
 {extends file="parent:frontend/checkout/change_payment.tpl"}
 
+{* Hidden input for retrieving the id of Mollie's Apple Pay payment mean *}
+{block name='frontend_checkout_payment_content'}
```
boxblinkracer, line 4: why is this important? please add comments, i think it might break something if removed?

PR #344, opened 2022-08-05
MOL-873: Enable more detailed payment errors

BlackScorp:
The Payment API and Order API can have more detailed error messages.

You can now enable the detailed error messages in plugin config.

The messages can be translated individually


Resources/config.xml
```diff
@@ -928,5 +928,60 @@
...
+
+        <element type="select" scope="shop" required="false">
+            <name>show_error_details</name>
+            <label lang="en">Show error details in checkout and payment</label>
+            <label lang="de">Fehlerdetails beim Checkout und bei der Zahlung anzeigen</label>
+            <label lang="nl-NL">Toon foutdetails bij afrekenen en betale</label>
+            <label lang="es-ES">Mostrar detalles del error en el proceso de compra y pago</label>
+            <label lang="fr-FR">Afficher les détails de l'erreur lors du paiement et du paiement</label>
+            <label lang="it-IT">Mostra i dettagli dell'errore in fase di pagamento e pagamento</label>
+            <value>inherited</value>
```
boxblinkracer, line 952: please change to NO as default

 <value>no</value>

Resources/config.xml
```diff
@@ -928,5 +928,60 @@
...
+            <label lang="it-IT">--------------------------------------------------------------------------------------- Errori API ---------------------------------------------------------------------------------------</label>
+            <options>
+                <position>30</position>
+            </options>
+        </element>
+
+        <element type="select" scope="shop" required="false">
+            <name>show_error_details</name>
+            <label lang="en">Show error details in checkout and payment</label>
+            <label lang="de">Fehlerdetails beim Checkout und bei der Zahlung anzeigen</label>
```
boxblinkracer, line 947: might sound annoying but we should stick to short and compact feature names.
can you change it to be that "german" title (+ other languages)
and also add a tooltip description as with the other features.

also,..please add it to the existing "order" section, that it totally fine because its the scope of ordering and checkout.

you need to adjust the positions of all following configs then :)

Shopware_5_7_13__Rev__202206221313__-_Backend__c__shopware_AG

Resources/services/services.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+
+
+    <services>
+        <service class="MollieShopware\Services\MollieOrderRequestAnonymizer\MollieOrderRequestAnonymizer" id="mollie_shopware.services.mollie_order_request_anonymizer.mollie_order_request_anonymizer" public="true">
```
boxblinkracer, line 10: could you improve the id please
seems very generated
try a simple "mollie_shopware.services.order_request_anonymizer",
thats totally fine
because of interfaces and options to have separate "beautiful" ids instead of shopware6, i try to create a good-named DI setup where i could easily switch implementation classes :)

Resources/services/services.xml
```diff
@@ -0,0 +1,19 @@
...
+            <argument type="string">***</argument>
+        </service>
+
+        <service class="MollieShopware\Services\Mollie\Payments\Extractor\PaymentFailedDetailExtractor" id="mollie_shopware.services.mollie.payments.extractor.payment_failed_detail_extractor" public="true"/>
```
boxblinkracer, line 14: maybe think about this one too, but doesnt look that redundant in the wordins as above

still, does it have to be public?

Gateways/Mollie/MollieGateway.php
```diff
@@ -151,12 +165,20 @@ public function createOrder(array $requestData)
         try {
             return $this->apiClient->orders->create($requestData);
         } catch (ApiException $ex) {
+            $anonymizedRequest = $this->mollieOrderAnonymizer->anonymize($requestData);
+            $this->logger->critical($ex->getMessage(), $anonymizedRequest);
```
boxblinkracer, line 169: i dont think this should be logged in here?
wasnt it before that the exception is passed on and then logged in the controller/action scope?
then we have double-logs here with the new implementation?

Gateways/Mollie/MollieGateway.php
```diff
@@ -151,12 +165,20 @@ public function createOrder(array $requestData)
...
+        }
+    }
+
+    public function createPayment(array $requestData)
+    {
+        try {
+            return $this->apiClient->payments->create($requestData);
+        } catch (ApiException $ex) {
+            $anonymizedRequest = $this->mollieOrderAnonymizer->anonymize($requestData);
+            $this->logger->critical($ex->getMessage(), $anonymizedRequest);
```
boxblinkracer, line 181: same thing here with the log?

Gateways/Mollie/MollieGateway.php
boxblinkracer: ok found it
so the concept is full logs on the outer scope
inner-scope is detailed logs in special use cases (like the invalid total amount that was done before).

problem, as seen on the screenshot, you've opened a new catch way in the controller action that doesnt contain a log. if one would remove inner logs, we have a way that is not logged at all anymore.

please add the log to the outer scope.
if it has to stick also in this gateway for the anonymous data, then we would indeed need 2 logs, and we could turn this one to DEBUG maybe? because it shows in-depth additional data, and make the outer-one the error one?

Screen Shot 2022-09-14 at 12 46 41

Tests/PHPUnit/Services/Mollie/Payments/Extractor/ApiExceptionDetailsExtractorTest.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        $exception = new ApiException('test', 1, 'test.field');
+
+        $result = $this->extractor->extractDetails($exception);
+        $this->assertNull($result);
+    }
+
+    public function testMessageWithoutDocumentation(): void
+    {
+        $exception = new ApiException('Error executing API call (123:Test): Test Message', 123, 'test.field');
```
boxblinkracer, line 42: could we have a separate test that explains us that the testfield will be extracted and added with an "Error" prefix in that object?
had a hard time figuring out

Controllers/Frontend/Mollie.php
```diff
@@ -186,28 +193,39 @@ public function directAction()
             }
 
             $this->redirect($session->getCheckoutUrl());
+        } catch (ApiException $ex) {
+            $paymentFailedDetails = $this->apiExceptionDetailsExtractor->extractDetails($ex);
+            if ($this->config->showDetailedErrorMessages() && $paymentFailedDetails !== null) {
+                $this->ERROR_PAYMENT_FAILED_REASON_CODE = $paymentFailedDetails->getReasonCode();
+                $this->ERROR_PAYMENT_FAILED_REASON_MESSAGE = $paymentFailedDetails->getReasonMessage();
+            }
+            $this->handleOnException();
```
boxblinkracer, line 202: log is missing as described in the gateway

boxblinkracer (changes requested):
see comments, thx

Gateways/Mollie/MollieGateway.php
BlackScorp: Ineed i had to log in inner scope because the data was availiable there, so i made a debug message in inner scope and error on outerscope without details

Tests/PHPUnit/Services/Mollie/Payments/Extractor/ApiExceptionDetailsExtractorTest.php
BlackScorp: i added here descriptions to the tests, i guess this is then more clear

PR #347, opened 2022-08-11
MOL-79 Added custom refund amount to order overview.

iamreinder:


Controllers/Backend/MollieOrders.php
```diff
@@ -334,14 +335,7 @@ public function refundAction()
...
 
-            # switch to the config and api key of the shop from the order
-            $shopSwitcher = new MollieShopSwitcher($this->container);
-            $this->config = $shopSwitcher->getConfig($order->getShop()->getId());
-            $this->apiClient = $shopSwitcher->getMollieApi($order->getShop()->getId());
-
-            $transaction = $this->orderService->getOrderTransactionByNumber($order->getNumber());
-
-            $refund = $this->refundService->refundFullOrder($order, $transaction);
+            $refund = $this->refundOrderFacade->refundOrder($order->getNumber(), $customAmount);
```
boxblinkracer, line 338: attention, you are removing the config and api switching!
in the backend, orders can come from any subshop...so this is always related to the provided order that will be refunded. i think this breaks things

boxblinkracer:
good job @iamreinder 
i would merge it

could you arrange the modal a bit before. 
the main (german) market has a truncated text...maybe you can increase the height, or even better, increase column1 a little bit. then it would all fit in 1 line

thank you

Screen Shot 2022-09-19 at 09 22 49

iamreinder:
@boxblinkracer I'm fixing the window right now, this is how it will look:

image

boxblinkracer:
hi @iamreinder 
awesome thanks
its still not completely visible (see textfield)
but i will just adjust that after the merge
thanks

PR #348, opened 2022-08-31
MOL-658 apple pay direct for esd products

iamreinder:


Tests/PHPUnit/Utils/Fakes/View/FakeView.php
```diff
@@ -36,7 +36,7 @@ public function assign($spec, $value = null, $nocache = null, $scope = null)
      */
     public function getAssign($spec = null)
     {
-        if (!$spec === null) {
+        if ($spec === null) {
```
boxblinkracer, line 39: sure?

Components/ApplePayDirect/Services/ApplePayButtonBuilder.php
```diff
@@ -50,20 +56,25 @@ class ApplePayButtonBuilder
...
+        $this->accountService = $accountService;
         $this->configMollie = $configMollie;
         $this->configShopware = $configShopware;
         $this->restrictionService = $restrictionService;
+        $this->applePayPaymentMethod = $applePayPaymentMethod;
 
         # attention, modules does not exist in CLI
-        $this->sAdmin = Shopware()->Modules()->Admin();
-        $this->applePayPaymentMethod = $applePayPaymentMethod;
+        $this->sAdmin = $sAdmin;
```
boxblinkracer, line 76: can you run the apple pay cli command and verify it does not crash, thanks

Components/ApplePayDirect/Services/ApplePayButtonBuilder.php
```diff
@@ -90,8 +100,17 @@ public function addButtonStatus(Enlight_Controller_Request_Request $request, Enl
             if ($isRiskManagementBlocked) {
                 $isActive = false;
             }
-        }
 
+            # if a customer has esd products in the basket, check if
+            # the customer is logged in with a full customer account
+            $hasEsdProductsInBasket = $controller === 'checkout' && $this->basketHasEsdProducts();
```
boxblinkracer, line 106: code looks working
but somehow i think we shouldnt keep it like this.
specific logic, inline strings...this seems to be at least a private function or maybe even moved to another class with unit tests

Resources/services/components/applepay_direct.xml
```diff
@@ -16,10 +16,13 @@
 
         <service id="mollie_shopware.components.apple_pay_direct.services.button_builder"
                  class="MollieShopware\Components\ApplePayDirect\Services\ApplePayButtonBuilder" public="true">
+            <argument type="service" id="mollie_shopware.components.account.account" />
             <argument type="service" id="mollie_shopware.config"/>
             <argument type="service" id="config"/>
             <argument type="service" id="mollie_shopware.components.apple_pay_direct.services.payment_method"/>
             <argument type="service" id="mollie_shopware.components.apple_pay_direct.services.display_option"/>
+            <argument type="expression">container.get('Shopware_Components_Modules').Admin()</argument>
```
boxblinkracer, line 24: sure that works in showare 5.2.21?

Tests/PHPUnit/Components/ApplePayDirect/Services/ApplePayButtonBuilderTest.php
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    /**
+     * @test
+     * @testdox Apple Pay direct button is not active when restricted by risk management.
+     *
+     * @throws Exception
+     * @return void
+     */
+    public function testIfApplePayButtonNotActiveWhenRestrictedByRiskManagement()
```
boxblinkracer, line 100: i really really like thos tests
still,  please find way way shorter names, it's nearly impossible to figure out existing cases when just looking at it, so it can indeed happen that a test exists twice

boxblinkracer (changes requested):
see comments

Tests/PHPUnit/Components/ApplePayDirect/Services/ApplePayButtonBuilderTest.php
```diff
@@ -0,0 +1,419 @@
...
+            ->willReturn($loggedIn);
+    }
+
+    /**
+     * Returns a preconfigured mock for the sAdmin module.
+     *
+     * @param string $countryIso
+     * @return sAdmin
+     */
+    private function getAdminModuleMock(string $countryIso = 'nl'): sAdmin
```
boxblinkracer, line 411: please no default values

Tests/PHPUnit/Utils/Fakes/View/FakeView.php
iamreinder: Yeah I am. Why would you want the full array if you specify a key? It's how the view in Shopware works as well. You get everything if you don't provide a key, but just one item if you do provide a key.

Components/ApplePayDirect/Services/ApplePayButtonBuilder.php
iamreinder: As you have seen further on, there are unit tests. But sure, I can put this in a separate class.

Resources/services/components/applepay_direct.xml
iamreinder: I'm not sure to be honest. It's something Shopware uses itself, so I thought it should be okay to use. I tried to pick the less worse option. The Shopware singleton isn't really the best way of doing things either. I'll check what works.

Tests/PHPUnit/Components/ApplePayDirect/Services/ApplePayButtonBuilderTest.php
iamreinder: I feel like being specific is quite important in tests. But I'll try to shorten them.

Tests/PHPUnit/Utils/Fakes/View/FakeView.php
boxblinkracer: i was just asking because u changed the condition (accidentally) :D

Tests/PHPUnit/Components/ApplePayDirect/Services/ApplePayButtonBuilderTest.php
boxblinkracer: and i didnt say sth different, specific and short if possible
tests need to cover both things, being specific and easy readable

Resources/services/components/applepay_direct.xml
boxblinkracer: could you please try, thank you

Components/ApplePayDirect/Services/ApplePayButtonBuilder.php
boxblinkracer: its okay like this thx

Tests/PHPUnit/Utils/Fakes/View/FakeView.php
iamreinder: Not accidentally ;-)

Resources/services/components/applepay_direct.xml
iamreinder: Well, you were right. Expressions do not work in Shopware 5.2.

I have a little trouble finding the right solution. Using the singleton works. It's also the best supported solution throughout many Shopware 5 version. But than the tests won't work, because the singleton isn't available without booting Shopware. I'm not sure what's the best way to deal with this. I have a solution, but **I would really like to talk it over before it is merged.**

Components/ApplePayDirect/Services/ApplePayButtonBuilder.php
iamreinder: There is no CLI command using this class. I didn't put the comment there, but I think it's meant as a warning in case there every would be a CLI command using the ApplePayButtonBuilder.

PR #352, opened 2022-09-22
MOL-896 fatal error on credit card service

iamreinder:
I added extra logging around retrieving the credit card payment token. There's an exception in the `CreditCardService` now that, when thrown, is catched in a try/catch block.

I believe the method could throw an exception, and thereby the checkout would fail, even if the token isn't required to finish the payment. To prevent that from happening, I catch the exception thrown by the `getCardToken` method (in `Components/Services/CreditCardService.php`) and log it.

### DIFF
--- a/Components/Services/PaymentService.php
+++ b/Components/Services/PaymentService.php
@@ -22,6 +22,7 @@ use MollieShopware\Components\Mollie\Services\TransactionUUID\TransactionUUID;
 use MollieShopware\Components\Mollie\Services\TransactionUUID\UnixTimestampGenerator;
 use MollieShopware\Components\MollieApiFactory;
 use MollieShopware\Components\Translation\Translation;
+use MollieShopware\Exceptions\CustomerNotFoundException;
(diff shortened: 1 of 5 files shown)


Components/Services/PaymentService.php
```diff
@@ -642,7 +649,16 @@ private function configurePaymentSettings(PaymentInterface $paymentMethodObject,
...
 
         if ($paymentMethodObject instanceof CreditCard) {
-            $ccToken = $this->creditCardService->getCardToken();
+            # prevent this method from failing, because at this
+            # point we're not sure if the credit card token
+            # is needed to finish the payment at Mollie
+            try {
+                $ccToken = $this->creditCardService->getCardToken();
+            } catch (CustomerNotFoundException $exception) {
+                $this->logger->error('Could not get the credit card token for the current customer.', [
```
boxblinkracer, line 658: could you change that to a warning.
we dont really know what this is.
the token is optional only, and merchants are sometimes scared that things break on errors.
warning would make it softer :)

PR #373, opened 2023-03-14
MOL-1105: Fix JPY rounding

BlackScorp:


Tests/PHPUnit/Services/Mollie/Payments/Formatters/NumberFormatterTest.php
```diff
@@ -34,8 +34,17 @@ public function testInteger()
     {
         $formatter = new NumberFormatter();
 
-        $value = $formatter->formatNumber(24);
+        $value = $formatter->formatNumber(24, 'eur');
 
         $this->assertEquals('24.00', $value);
     }
+
+    public function testJPYRounding()
```
boxblinkracer, line 42: can you add a comment that this is necessary because JPY has no decimals?

PR #394, opened 2024-01-23
NTR: fix select first language in admin login

BlackScorp:


Tests/Cypress/cypress/support/actions/backend/ConfigSetupAction.js
```diff
@@ -60,7 +60,7 @@ export default class ConfigSetupAction {
         repoLogin.getEmail().clear().type('demo');
         repoLogin.getPassword().clear().type('demo');
         repoLogin.getLocale().click();
-        cy.contains('German (Germany)').click();
+        cy.get('.x-boundlist-item:first').click();
```
boxblinkracer, line 63: attention, sometimes english was the first (i think)
thats why german was explicitely used

boxblinkracer (changes requested):
comments

